PR #9, opened 2021-12-27
feat: identities genSecret multipart secret updates & test updates

bdim1:
- Identities generate secret function updated
- Test updates to reflect the changes


packages/identity/src/identity.ts
```diff
@@ -75,42 +78,35 @@ class ZkIdentity {
...
-   * Generate commitment from identity secret
-   * @returns identity commitment
+   * Creates secret from ZkIdentity
+   * @returns
    */
-  genIdentityCommitmentFromSecret(): bigint {
-    if (!this.secret.length) throw new Error("Secret is not generated")
-    const secretHash = poseidonHash(this.secret)
-    return poseidonHash([secretHash])
+  genSecret(parts: number = 2): void {
```
akinovak, line 91: I guess we always want to make initial secret = trapdor ** nullifier and then make SPAM_THRESHOLD degrees of that initial value?

akinovak (approved):
lgtm!

Issue #32, opened 2022-10-01
IncrementalBinaryMerkleTree.sol: Update can change uninitialized leaves in the tree


twister-dev:
Although this is a description for IncrementalBinaryMerkleTree.sol, it probably applies to the Quin Tree contract as well.

**Describe the bug**

Long story short: It is possible to use the `update` function to change a zero leaf at an index that hasn't been inserted into the tree yet.
 
For example, suppose that 4 leaves are inserted into the tree. We can update leaf 7 by proving inclusion of the zero value at leaf 7, because technically the tree is never "empty", but it begins completely filled with zero values. 

Updating an uninitialized index can cause the tree root to no longer represent the set of leaves accurately, because the lastSubtrees array will be incorrect at some point when updating the root.

**To Reproduce**
You can see a demonstration of this bug in a unit test I wrote on my fork of the package:

[IncrementalBinaryTreeTest.ts#L188](https://github.com/twister-dev/zk-kit/blob/8f596f17605907cd84aa284abf69969729836743/packages/incremental-merkle-tree.sol/test/IncrementalBinaryTreeTest.ts#L188)

**Expected behavior**
The contract should reject updates that reach an index greater than or equal to the number of leaves in the tree. I implemented this in the next commit of the same fork and branch:

contract: [IncrementalBinaryTree.sol#L124](https://github.com/twister-dev/zk-kit/blob/b66b24628c966b242be158956507684a9831635b/packages/incremental-merkle-tree.sol/contracts/IncrementalBinaryTree.sol#L124)
test: [IncrementalBinaryTreeTest.ts#L188](https://github.com/twister-dev/zk-kit/blob/b66b24628c966b242be158956507684a9831635b/packages/incremental-merkle-tree.sol/test/IncrementalBinaryTreeTest.ts#L188)

**Additional context**
I believe I patched the bug in the IncrementalBinaryMerkleTree.sol contract, as shared above, and if it is a satisfiable solution then I can submit a pull request with that change.

However, I'm mainly opening this bug report because I haven't studied Quin Merkle Trees and don't know how to patch that contract (if the bug is even there, but I'm assuming it is).

cedoor:
Hi @twister-dev , thank you very much for the comprehensive description of this bug. I think your solution is right, and if you can also submit a PR, that would be great!

The Quin Merkle tree is exactly like a Binary Merkle tree, but with 5 child nodes per node, so the bug should be the same there. But if you're not sure we can fix it.

[...]

PR #33, opened 2022-10-01
Dev

twister-dev:
## Description
Change the IncrementalBinaryTree.sol `update` method to require the index of the leaf being updated to be already inserted into the tree.

We do this by recovering the `updateIndex` from the `proofPathIndices` array, then comparing with `self.numberOfLeaves` in the tree. Because `self.numberOfLeaves` is the length of the leaves in the tree, which is one greater than the index of the last leaf in the tree, the requirement that `updateIndex < self.numberOfLeaves` covers all the existing leaves in the tree.

I also added a unit test to check that this operation reverts when an uninitialized leaf is updated. [IncrementalBinaryTreeTest.ts#L188](https://github.com/twister-dev/zk-kit/blob/dev/packages/incremental-merkle-tree.sol/test/IncrementalBinaryTreeTest.ts#L188)

## Related Issue

This is a bug fix as described in [zk-kit/issues/32](https://github.com/privacy-scaling-explorations/zk-kit/issues/32).

-   [ ] Yes
-   [X] No

## Other information

The IncrementalQuinTree requires a bug fix, too. I am looking into fixing it as well.

### DIFF
--- a/packages/incremental-merkle-tree.sol/contracts/IncrementalBinaryTree.sol
+++ b/packages/incremental-merkle-tree.sol/contracts/IncrementalBinaryTree.sol
@@ -102,7 +102,7 @@ library IncrementalBinaryTree {
 
         uint256 updateIndex;
         for (uint8 i = 0; i < depth; ) {
-            updateIndex |= uint(proofPathIndices[i] & 1) << uint(i);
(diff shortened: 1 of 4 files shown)


cedoor:
Thanks @twister-dev, great job! LGTM

twister-dev:
Thanks! It's an honor to contribute to such a foundational tool.

PR #43, opened 2023-03-21
feat(incremental-merkle-tree.sol): support default zeroes as constants

### LINKED ISSUE #42: Store a default `zeroes` value for incremental merkle trees
In the unirep project we need to create many trees over time. To do this we have to copy the `zeroes` mapping to each new `IncrementalTreeData` structure on initialization. This involves `depth` `SLOAD` and `SSTORE` operations.

Store an optional `defaultZeroes` for the `IncrementalBinaryTree` library. Add a flag to the `IncrementalTreeData` `defaultZeroes` that will cause the implementation to use the zeroes included in the library.

Rolling a modified version of the incremental merkle tree ourselves.

**Additional context**

vimwitch:
## Description

This PR adds pre-computed constant zero leaves as constants in the `IncrementalBinaryTree.sol` library. It also adds an option (default `false`) to `useDefaultZeroes`. Such trees can be initialized using `initDefaultZeroes()` which does not involve doing `depth` poseidon hashes/`SSTORE`.

This change should be non-breaking. Existing implementations will default to having `useDefaultZeroes = false` which should not change behavior.

Tests are added for `insert` and `remove` operations on default zero trees. Other operations are unchanged.

## Related Issue

Closes #42 

-   [ ] Yes
-   [x] No

### DIFF
--- a/packages/incremental-merkle-tree.sol/contracts/IncrementalBinaryTree.sol
+++ b/packages/incremental-merkle-tree.sol/contracts/IncrementalBinaryTree.sol
@@ -12,6 +12,7 @@ struct IncrementalTreeData {
     mapping(uint256 => uint256) zeroes; // Zero hashes used for empty nodes (level -> zero hash).
     // The nodes of the subtrees used in the last addition of a leaf (level -> [left node, right node]).
     mapping(uint256 => uint256[2]) lastSubtrees; // Caching these values is essential to efficient appends.
+    bool useDefaultZeroes;
 }
 
 /// @title Incremental binary Merkle tree.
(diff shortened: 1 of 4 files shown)


packages/incremental-merkle-tree.sol/contracts/IncrementalBinaryTree.sol
```diff
@@ -48,6 +120,17 @@ library IncrementalBinaryTree {
         self.root = zero;
     }
 
+    function initDefaultZeroes(IncrementalTreeData storage self, uint256 depth) public {
```
cedoor, line 123: What about `initWithDefaultZeroes`? `initDefaultZeroes` suggests you are initializing the default zeroes (which are actually constants), and not the tree.

cedoor (approved):

Issue #48, opened 2023-03-29
Proposal: add a new solidity package for HashTower


LCamel:
I would like to contribute a data structure to the project: HashTower.

The purpose of it is similar to Incremental Binary Tree.

The amortized cost for adding an item into it is O(1).

REF: https://github.com/LCamel/HashTower/tree/85bff67f73c1893d6cbcf1b093f69d4e7e2dfbdb

PR #49, opened 2023-03-29
Hashtower init

LCamel:
## Description

The initial version for the package hashtower.sol .

## Related Issue

#48 

-   [ ] Yes
-   [X] No


packages/hashtower.sol/package.json
```diff
@@ -0,0 +1,50 @@
+{
+    "name": "hashtower.sol",
```
cedoor, line 2: Is it ok for you to use `@zk-kit/hashtower.sol` as a package name?

packages/hashtower.sol/package.json
```diff
@@ -0,0 +1,50 @@
...
+        "hardhat-gas-reporter": "^1.0.8",
+        "prettier-plugin-solidity": "^1.0.0-beta.19",
+        "solhint": "^3.3.6",
+        "solhint-plugin-prettier": "^0.0.5",
+        "solidity-coverage": "^0.8.1",
+        "typechain": "^8.1.0"
+    },
+    "dependencies": {
+        "poseidon-solidity": "0.0.3"
+    },
```
cedoor, line 38: Would you like to add the authors of this package?

cedoor (commented):
Hey @LCamel, this is great! Thank you so much for your PR. Would you like to add a README file with credits?

Also, which version should we publish?

packages/hashtower.sol/package.json
LCamel: I just added the missing `contracts/package.json` with `"name": "@zk-kit/hashtower.sol"` in it.

packages/hashtower.sol/package.json
LCamel: Added. (in `contracts/package.json`)

packages/hashtower.sol/contracts/HashTower.sol
```diff
@@ -0,0 +1,120 @@
...
+library HashTower {
+    uint256 internal constant SNARK_SCALAR_FIELD =
+        21888242871839275222246405745257275088548364400416034343698204186575808495617;
+
+    event Add(uint8 indexed level, uint64 indexed lvFullIndex, uint256 value);
+
+    /// @dev Add an item.
+    /// @param self: HashTower data
+    /// @param item: item to be added
+    function add(HashTowerData storage self, uint256 item) public {
```
vimwitch, line 31: This implementation uses the double nested approach where each level is a hashchain like `H(H(H(item0, item1), item2), item3)`. This implementation is very unfavorable in zk. For example, if i want to prove that `item1` exists in `R = H(H(H(item0, item1), item2), item3)` I need to provide `item0`, `item1`, `item2` and `item3`. This leads to 3x `PoseidonT3` inside the proof.

If I'm proving against a 4x24 structure each proof will require 72 `PoseidonT3`. I think this tradeoff is not worth it when the same proof could be done using 24 `PoseidonT5` if each level is simply `H(item0, item1, item2, item3)`.

The tradeoff is increased gas cost per insertion. In solidity the poseidon function scales slightly non-linearly with input count, `PoseidonT5` is 2.2x more expensive than `PoseidonT3` despite having only 2x inputs. Regardless minimizing constraints is key to building user applications.

I think in the future we could add an option to recursively hash each level, but for now we need to choose one or the other, and I'd rather go with the more zk friendly option. Also bear in mind we'll likely find more optimizations like lazy digest calculation.

packages/hashtower.sol/contracts/HashTower.sol
```diff
@@ -0,0 +1,120 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.4;
+
+import {PoseidonT3} from "poseidon-solidity/PoseidonT3.sol";
+
+// CAPACITY = W * (W**0 + W**1 + ... + W**(H - 1)) = W * (W**H - 1) / (W - 1)
+uint256 constant H = 24;
+uint256 constant W = 4;
+uint256 constant CAPACITY = 375_299_968_947_540;
```
vimwitch, line 9: These should not be constants and should be set by the user. Definitely allow a variable `H`, _maybe_ allow a variable `W` within a range (maybe like `[2-6]`).

packages/hashtower.sol/contracts/HashTower.sol
vimwitch: So reading more the `digestOfDigests` is also a hashchain so the zk scaling should be worse than I wrote above I think? I'd much rather we use the polysum to calculate the digest of the structure itself. Using a constant `R` value is unsafe, but perhaps if we sample an `R**i` value based on the contents of the structure this can be more safe.

Something like take `R**i = levelHash[i-1]` (where `levelHash[-1] = 1`) might work and would keep the favorable gas/constraints property of the polysum.

Even if we don't use a polysum we can form a tree of high arity poseidon hashes. To get `H=24` we can do a `PoseidonT5` hash of `PoseidonT6` hashes. e.g. a shallow tree. This would cost (`PoseidonT5 + PoseidonT6` gas `159k`) (plus the constraints in zk).

vimwitch (changes requested):
Implementation of this version looks good, I think we should just change the structure

packages/hashtower.sol/contracts/HashTower.sol
LCamel: I think we may choose a large enough `H` in the contract for the library users without incurring costs:
- the storage slots at the high levels will remain 0 before being used
- library users can pick their own `H` in the circuit independent of the `H` in the contract
- as the count of items growths, library users can upgrade the circuit without changing the storage

LCamel: For `W`, it does lead to a tradeoff between the circuit complexity and the on-chain cost.
Increasing W will linearly increase the circuit complexity, and decrease the on-chain cost a little bit.
I have done an outdated calculation before. I should do it for the new contract / circuit.

I think there are some possibilities:

1. Fixed W
1.1 Choose a fixed W for the library users
1.2 Deploy different copies for W=[2-6]
2. Dynamic W
2.1 Merge W with countOfItems, so it will not incur an additional SLOAD

LCamel: For `W`, it does lead to a tradeoff between the circuit complexity and the on-chain cost.
Increasing W will linearly increase the circuit complexity, and decrease the on-chain cost a little bit.
I have done an outdated calculation before. I should do it for the new contract / circuit.

I think there are some possibilities:

1. Static W
1.1 Choose a fixed W for the library users
1.2 Deploy different copies for W=[2-6]
2. Dynamic W
2.1 Merge W with countOfItems, so it will not incur an additional SLOAD

packages/hashtower.sol/contracts/HashTower.sol
LCamel: > If I'm proving against a 4x24 structure each proof will require 72 PoseidonT3. I think this tradeoff is not worth it when the same proof could be done using 24 PoseidonT5 if each level is simply H(item0, item1, item2, item3).

Does that require 4 on-chain PoseidonT5 for the 4 items...?

> The tradeoff is increased gas cost per insertion. In solidity the poseidon function scales slightly non-linearly with input count, PoseidonT5 is 2.2x more expensive than PoseidonT3 despite having only 2x inputs. Regardless minimizing constraints is key to building user applications.

I would like to minimize the gas cost while keeping an acceptable circuit complexity.  However,  I'm not quite familiar with how large the circuit will become unacceptable. Are 10000~20000 constraints acceptable ...?

| W = 4, H = | capacity        | non-linear constraints | zkey size |
|------------|-----------------|------------------------|-----------|
|         12 |        22369620 |                  11738 |   6858502 |
|         16 |      5726623060 |                  15786 |   8860294 |
|         20 |   1466015503700 |                  19866 |  11925126 |
|         24 | 375299968947540 |                  23978 |  13955846 |

> I think in the future we could add an option to recursively hash each level, but for now we need to choose one or the other, and I'd rather go with the more zk friendly option. Also bear in mind we'll likely find more optimizations like lazy digest calculation.

I think both digest() and digestOfDigest() should have much room for improvement. I would like to make them pluggable like the Javascript version, but they are now hardcoded/inlined in the Solidity version for saving gas.
When there are better solutions, we can push a new one.

May we take this 0.0.1 version as the baseline for improvement?

packages/hashtower.sol/package.json
cedoor: Yes, it makes sense

packages/hashtower.sol/contracts/HashTower.sol
vimwitch: > Does that require 4 on-chain PoseidonT5 for the 4 items...?
No, it requires 1 `PoseidonT5`. The 4 items will be an input this.

> I would like to minimize the gas cost while keeping an acceptable circuit complexity. However, I'm not quite familiar with how large the circuit will become unacceptable. Are 10000~20000 constraints acceptable ...?

It's higher than ideal, remember this is just one component in the circuit. So we start with 11k constraints, then add more logic on top. We generally want to stay below 20k constraints total. Also bear in mind some circuits need multiple inclusion proofs.

> May we take this 0.0.1 version as the baseline for improvement?

I think we can merge this into `dev` and not publish to npm to continue working on it.

vimwitch (approved):
@cedoor what do you think of merging this without publishing to npm and working on the `dev` branch?

cedoor:
> @cedoor what do you think of merging this without publishing to npm and working on the `dev` branch?

Yes, it makes sense 👍🏽 Working on the `dev` branch by merging new PRs?

PR #50, opened 2023-04-01
Hashtower refactor

vimwitch:
## Description

Makes the hash tower less recursive. Instead use `H` levels of `W` elements. Each level digest being a single poseidon hash.

Implements the stronger `R` values that is a function of tower items.

-   [x] Yes
-   [ ] No


vimwitch:
Depends #49 

Average insertion gas cost for 500 items: `193k`

@LCamel Take a look and lmk what you think. I'm pretty sure we can figure out a way to use the polyum function for each level by trading incrementality for stronger `R` values.

vimwitch:
Something else we should consider, this type of structure is advantageous against a traditional tree because the number of elements scales up more aggressively.

e.g. comparing a binary tree and a hashtower with `W = 2`
```
hashtower depth 10: 2046 items
tree depth 10: 1024 items

hashtower depth 20: 2,097,150 items
tree depth 20: 1,048,576 items

hashtower depth 30: 2,147,483,646 items
tree depth 30: 1,073,741,824 items
```

So for the same constraints in ZK we get double the number of items with a hashtower as compared to a binary tree.

vimwitch:
Ok so i think we can use the polysum for the whole data structure. I'm using an approach in unirep where I take a random value, polysum it, hash the polysum output, then use that output as the `R` in a second polysum.

So if I have `data[]` that i want to checksum I'll take a random value `r` and do

```js
const F = // field constant
const r = // strong random value that is externally verified
const data = [] // the data to hash

let R = 0n;

for (let x = 0; x < data.length; x++) {
  R = (R + data[x] * r**x) % F
}

R = poseidon1([R])

// now do the final polysum

let checksum = 0

for (let x = 0; x < data.length; x++) {
  checksum = (checksum + data[x] * R**x) % F
}

return checksum
```

Imagining the case of a single item, to find a collision we need an `item` and `item1` such that

```js
poseidon1([item]) * item == poseidon1([item1]) * item1
```

This _should_ be pretty close to the collision resistance of poseidon.

Applying it to the data structure, we could choose a constant `r` and do the digest for each level like described above. Each insertion would cost 1 `PoseidonT2` for each level that is changed. In ZK we would only need `H` `PoseidonT2` operations.

cedoor (approved):

vimwitch:
Going to close this, the branch is still present in my fork

Issue #51, opened 2023-04-10
Refactor HashTower: emit leaf only and use packed level lengths


LCamel:
**Is your feature request related to a problem? Please describe.**

The current HashTower implementation emits events of internal nodes.
It is prepared for downloading only O(log N) events to build a proof.
However, it seems that partially downloading events might be unsafe:
https://discord.com/channels/943612659163602974/955585525081837628/1092647984887504907
So we go back to emit only the leaf events. 

**Describe the solution you'd like**

Internal nodes will no longer emit events.
Full level length related code can be removed.
We can pack level lengths in to a single uint256.
By removing the count -> length logic, the code might be easier to read.
The gas will also be reduced.

**Describe alternatives you've considered**

**Additional context**

https://discord.com/channels/943612659163602974/1065263015622090802/1093713774088835182

PR #52, opened 2023-04-10
Hashtower: emit leaf only and use packed level lengths (and rename)

LCamel:
## Description

Internal nodes will no longer emit events.
Full level length related code can be removed.
We can pack level lengths in to a single uint256.
By removing the count -> length logic, the code might be easier to read.
The gas will also be reduced.

## Related Issue

#51 

-   [X] Yes
-   [ ] No

Currently there is no user for the library, so there will be no user impact.

## Other information

The specific implementation is also renamed to HashTowerHashChain, so we may have different implementations for different tradeoffs.


vimwitch (approved):
This looks good to me.

cedoor (approved):

Issue #53, opened 2023-04-18
Add the Javascript package @zk-kit/hashtower


LCamel:
**Describe the package you'd like**
Add the Javascript package `@zk-kit/hashtower` .

**Additional context**
Developers can use this package to build proofs for `@zk-kit/hashtower.sol` .

PR #54, opened 2023-04-18
Hashtower js

LCamel:
## Description

Add the Javascript package `@zk-kit/hashtower` .
Developers can use this package to build proofs for `@zk-kit/hashtower.sol` .

## Related Issue

#53 

-   [ ] Yes
-   [X] No


packages/hashtower/src/hashtower-hash-chain-proof-builder.ts
```diff
@@ -0,0 +1,99 @@
...
+const pad0 = (arr: any, len: number) => pad(arr, len, BigInt(0))
+const pad00 = (arr2D: any, h: number, w: number) => pad(arr2D, h, []).map((a: any) => pad0(a, w))
+
+/**
+ * HashTowerHashChainProofBuilder is a TypeScript implementation of HashTower to generate proofs of membership.
+ * @param H Height of tower of the proving circuit. It can be less than the H in the contract.
+ * @param W Width of tower.
+ * @param hash A hash function which supports 2 input values.
+ */
+export default function HashTowerHashChainProofBuilder(H: number, W: number, hash: HashFunction) {
```
cedoor, line 14: What about a JS class here?

cedoor: I can't remember the specific reason but using separate functions made the algorithm faster. I think there were no differences between function and class tho. That code should be compiled into JS functions anyway in theory

LCamel: I did some benchmarking, and there was no significant difference.

I pushed a new TS-class commit.

packages/hashtower/src/add.ts
```diff
@@ -0,0 +1,34 @@
+import checkParameter from "./checkParameter"
+import { HashFunction } from "./types"
+
+export default function _add(
```
vimwitch, line 4: I'm _really_ not a fan of this separate file, separate function design. It's super hard to follow what is happening because we're passing properties from the main instance by reference and then mutating them.

I'd much rather see 1 file that is 200 lines than 5 files each < 50 lines.

packages/hashtower/src/add.ts
```diff
@@ -0,0 +1,34 @@
...
+    item: BigInt,
+    H: number,
+    W: number,
+    digestFunc: HashFunction,
+    levels: BigInt[][],
```
vimwitch, line 9: Prefer the `bigint` type

packages/hashtower/src/types/index.ts
```diff
@@ -0,0 +1,11 @@
+export type HashFunction = (BigInts: BigInt[]) => BigInt
```
vimwitch, line 1: Can these types be inlined/exported in the main class file?

packages/hashtower/src/hashtower-hash-chain-proof-builder.ts
```diff
@@ -0,0 +1,57 @@
...
+    private readonly _levels: BigInt[][] = []
+    private readonly _fullLevels: BigInt[][] = []
+
+    /**
+     * Initializes the proof builder.
+     * @param H Height of tower of the proving circuit. It can be less than the height declared in the contract.
+     * @param W Width of tower.
+     * @param hash A hash function which supports 2 values.
+     */
+    constructor(H: number, W: number, hash: HashFunction) {
```
vimwitch, line 23: I think `hash` should have a default value like `poseidon2`

packages/hashtower/src/hashtower-hash-chain-proof-builder.ts
```diff
@@ -0,0 +1,57 @@
...
+import _indexOf from "./indexOf"
+import _build from "./build"
+
+/**
+ * HashTowerHashChainProofBuilder is a TypeScript implementation of HashTower to generate proofs of membership.
+ */
+export default class HashTowerHashChainProofBuilder {
+    private readonly _H: number
+    private readonly _W: number
+    private readonly _digestFunc: HashFunction
```
vimwitch, line 13: I'd almost like this `HashFunction` type to be replaced with `(bigint, bigint) => bigint`. I really dislike custom types that are _super_ thin wrappers. It just adds more things I have to figure out.

`(bigint, bigint) => bigint` can be understood at a glance.

This is made 10x worse when the type definitions are in a different file.

vimwitch (changes requested):
I really dislike having lots of files and complicated types. Having a single class in a single file with primitive types is _much_ easier to understand because it follows a pattern used by many other languages/implementations and can be read without switching files.

packages/hashtower/src/add.ts
cedoor: Re 1 file with 200 lines vs 4 with 50 lines, I agree in general, but I remember there were some differences in terms of performance. I need to check it again

LCamel: I think most of the computation is on poseidon(), the data structure itself consume very little time.

However, the time of poseidon() itself varies about 10% on my computer between different runs.

```javascript
import { poseidon } from "circomlibjs";

console.time("1")
for (let i = 0; i < 10000; i++) { poseidon([i, i + 1]) }
console.timeEnd("1")

console.time("1")
for (let i = 0; i < 10000; i++) { poseidon([i, i + 1]) }
console.timeEnd("1")
```

```shell
hashtower % node a.mjs
1: 5.573s
1: 5.611s
hashtower % node a.mjs
1: 5.751s
1: 5.820s
hashtower % node a.mjs
1: 5.735s
1: 5.969s
hashtower % node a.mjs
1: 5.862s
1: 5.739s
hashtower % node a.mjs
1: 6.054s
1: 5.953s
```

vimwitch: You should try benchmarking `poseidon-lite`, it might be faster. In the latest version i removed some of the modulo reduction to trade memory for compute. I was seeing < 1 ms / hash even using implementation without pre-multiplied constants.

LCamel: `poseidon-lite` is about 50% faster.
https://gist.github.com/LCamel/f9f1c377af8dd6aba48f2c68ae9ce80e

packages/hashtower/src/hashtower-hash-chain-proof-builder.ts
LCamel: Replaced.

packages/hashtower/src/types/index.ts
LCamel: Inlined.

packages/hashtower/src/add.ts
LCamel: Replaced.

vimwitch (approved):

PR #59, opened 2023-05-08
New parameter to init tree with a list of leaves

cedoor:
## Description

This PR adds a new parameter to the `IncrementalMerkleTree` JS class to allow devs to efficiently add a list of leaves without the need to add them manually with the `insert` function and then avoiding calculating the same hashes/non-leaf nodes a number of times equal to the tree arity.

-   [ ] Yes
-   [x] No

### DIFF
--- a/packages/incremental-merkle-tree/src/incremental-merkle-tree.ts
+++ b/packages/incremental-merkle-tree/src/incremental-merkle-tree.ts
@@ -16,7 +16,6 @@ import _verifyProof from "./verifyProof"
 export default class IncrementalMerkleTree {
     static readonly maxDepth = 32
 
-    private _root: Node
     private readonly _nodes: Node[][]
     private readonly _zeroes: Node[]
     private readonly _hash: HashFunction
@@ -30,15 +29,21 @@ export default class IncrementalMerkleTree {
      * @param depth Tree depth.
      * @param zeroValue Zero values for zeroes.
      * @param arity The number of children for each node.
+     * @param leaves The list of initial leaves.
      */
-    constructor(hash: HashFunction, depth: number, zeroValue: Node, arity = 2) {
+    constructor(hash: HashFunction, depth: number, zeroValue: Node, arity = 2, leaves: Node[] = []) {
         checkParameter(hash, "hash", "function")
         checkParameter(depth, "depth", "number")
         checkParameter(zeroValue, "zeroValue", "number", "string", "bigint")
         checkParameter(arity, "arity", "number")
+        checkParameter(leaves, "leaves", "object")
 
         if (depth < 1 || depth > IncrementalMerkleTree.maxDepth) {
             throw new Error("The tree depth must be between 1 and 32")
+        }
+
+        if (leaves.length > arity ** depth) {
+            throw new Error(`The tree cannot contain more than ${arity ** depth} leaves`)
         }
 
         // Initialize the attributes.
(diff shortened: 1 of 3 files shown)


packages/incremental-merkle-tree/src/incremental-merkle-tree.ts
```diff
@@ -30,33 +29,59 @@ export default class IncrementalMerkleTree {
      * @param depth Tree depth.
      * @param zeroValue Zero values for zeroes.
      * @param arity The number of children for each node.
+     * @param leaves The list of initial leaves.
      */
-    constructor(hash: HashFunction, depth: number, zeroValue: Node, arity = 2) {
+    constructor(hash: HashFunction, depth: number, zeroValue: Node, arity = 2, leaves: Node[] = []) {
```
ichub, line 34: yes, adding `leaves` as a parameter here is exactly what is needed!

packages/incremental-merkle-tree/src/incremental-merkle-tree.ts
```diff
@@ -30,33 +29,59 @@ export default class IncrementalMerkleTree {
...
-        // The default root is the last zero value.
-        this._root = zeroValue
+        this._nodes[depth] = []
+
+        // It initializes the tree with a list of leaves if there are any.
+        if (leaves.length > 0) {
+            this._nodes[0] = leaves
+
+            for (let level = 0; level < depth; level += 1) {
+                for (let index = 0; index < Math.ceil(this._nodes[level].length / arity); index += 1) {
```
vimwitch, line 70: This can be `Math.floor` because you're iterating forward in the internal loop.

packages/incremental-merkle-tree/src/incremental-merkle-tree.ts
```diff
@@ -30,33 +29,59 @@ export default class IncrementalMerkleTree {
...
+                    const position = index * arity
+                    const children = []
+
+                    for (let i = position; i < position + arity; i += 1) {
+                        children.push(this._nodes[level][i] || this.zeroes[level])
```
vimwitch, line 75: There is a bug here if the default leaf is non-zero and the initial leaves array contains a `0`. If `this._nodes[level][i] == 0` then we get `this.zeroes[level]` which could be non-zero. This is because `0` is falsey.

To fix this use the null coalescing operator `??`

packages/incremental-merkle-tree/src/incremental-merkle-tree.ts
```diff
@@ -30,33 +29,59 @@ export default class IncrementalMerkleTree {
...
+        // It initializes the tree with a list of leaves if there are any.
+        if (leaves.length > 0) {
+            this._nodes[0] = leaves
+
+            for (let level = 0; level < depth; level += 1) {
+                for (let index = 0; index < Math.ceil(this._nodes[level].length / arity); index += 1) {
+                    const position = index * arity
+                    const children = []
+
+                    for (let i = position; i < position + arity; i += 1) {
```
vimwitch, line 74: Bit of a nit pick, but it would be more clear that we're iterating over the arity if it were written like

```js
for (let i = 0; i < arity; i++) {
  children.push(this._nodes[level][position + i] ?? this.zeroes[level])
}
```

packages/incremental-merkle-tree/tests/index.test.ts
```diff
@@ -30,13 +32,39 @@ describe("Incremental Merkle Tree", () => {
...
+
             it("Should initialize a tree", () => {
                 expect(tree.depth).toEqual(depth)
                 expect(tree.leaves).toHaveLength(0)
                 expect(tree.zeroes).toHaveLength(depth)
                 expect(tree.arity).toEqual(arity)
             })
 
+            it("Should initialize a tree with 100 leaves", () => {
+                const leaves = Array.from(Array(100).keys())
```
vimwitch, line 51: You should test variable arity trees with variable number of leaves. Or at least test even and odd numbers of leaves for a binary tree

e.g.

```js
for (let arity = 0; arity < 16; arity++) {
  for (let leafMod = 0; leafMod < 16; leafMod++) {
    const leafCount = 100 + leafMod
    const leaves = Array(leafCount).fill().map((_, i) => i)
    const tree = new IncrementalMerkleTree(poseidon, depth, 0n, arity, leaves)
    // ... build second tree and compare
  }
}
```

packages/incremental-merkle-tree/src/incremental-merkle-tree.ts
cedoor: It needs `Math.ceil` here because it must cycle over an even number of nodes at each level. Otherwise it wouldn't hash the last node.

packages/incremental-merkle-tree/src/incremental-merkle-tree.ts
cedoor: Oh good catch! Thanks 👍🏽

packages/incremental-merkle-tree/src/incremental-merkle-tree.ts
cedoor: THat's more clear yea! Thanks

packages/incremental-merkle-tree/src/incremental-merkle-tree.ts
vimwitch: Ah no it must be `Math.ceil` otherwise we skip elements, you're right

PR #64, opened 2023-06-02
Calculate root for variable heights

vimwitch:
## Description

Overloads the `root` function in `LazyMerkleTree` to accept an optional depth.

-   [ ] Yes
-   [x] No

### DIFF
--- a/packages/incremental-merkle-tree.sol/contracts/LazyMerkleTree.sol
+++ b/packages/incremental-merkle-tree.sol/contracts/LazyMerkleTree.sol
@@ -93,7 +93,7 @@ library LazyMerkleTree {
 
     function root(LazyTreeData storage self, uint8 depth) public view returns (uint256) {
         uint40 numberOfLeaves = self.numberOfLeaves;
-        require(2**depth > numberOfLeaves, "LazyMerkleTree: ambiguous depth");
+        require(2**depth >= numberOfLeaves, "LazyMerkleTree: ambiguous depth");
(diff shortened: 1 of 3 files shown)


cedoor (approved):

packages/incremental-merkle-tree.sol/contracts/LazyMerkleTree.sol
```diff
@@ -82,9 +82,30 @@ library LazyMerkleTree {
...
 
     function root(LazyTreeData storage self) public view returns (uint256) {
-        // this will always short circuit if self.numberOfLeaves == 0
         uint40 numberOfLeaves = self.numberOfLeaves;
-        if (numberOfLeaves == 0) return defaultZero(1);
+        // dynamically determine a depth
+        uint8 depth = 1;
+        while (uint8(2)**depth < numberOfLeaves) {
+            depth++;
+        }
```
cedoor, line 90: You can calculate this in O(1) I think: `Math.ceil(Math.log2(numberOfLeaves))` in JS.

vimwitch: Ah yes, that was my original approach but I don't think there's a cheap way to implement that in solidity.

cedoor: Right!

cedoor (approved):

Issue #89, opened 2023-12-02
Add Noir SMT BN implementation


fabianschu:
**Is your feature request related to a problem? Please describe.**
@zk-kit/smt contains a great SMT implementation in Javascript. It would be neat to have a corresponding Noir circuit at hand to be able to verify (non-)membership proofs and it could enable further applications being able to manipulate SMTs from within a circuit (just like you can in the JS implementation)

**Describe the solution you'd like**
A Noir library with functions for `verify`, `add`, `update` and `delete` for BN-based SMTs (based on @zk-kit/smt implementation).

**Describe alternatives you've considered**
-

**Additional context**
I worked on respective circuits and discussed with @cedoor that this repo might be a good place for them to live here: https://github.com/privacy-scaling-explorations/zk-kit/issues/73#issuecomment-1830295070

cedoor:
Hey @fabianschu, thank you! Feel free to open a PR 👍🏽 

Do you feel like `@zk-kit/circuits` is a good place? Or does Noir need many configurations and so a separate package would be better?

PR #90, opened 2023-12-02
Adds Noir library for BN SMT `verify`, `add`, `update`, `delete`

fabianschu:
## Description

Includes a Noir circuit library based on the BigNumber smt JS implementation in `packages/smt`. The library includes functions for:
- verifying (non-)membership proofs
- adding a new entry
- deleting an existing entry
- updating an existing entry

Since it's written as a library and intended to be used in other circuits, I haven't added compilation scripts. More info about Noir libraries [here (had to use the webarchive because noir docs are down atm^^) ](https://web.archive.org/web/20231021012840/https://noir-lang.org/modules_packages_crates/crates_and_packages#libraries).

## Related Issue

#89 

-   [ ] Yes
-   [x] No


cedoor:
@fabianschu Thank you very much 🙏🏽

Issue #66, opened 2023-09-09
Proposal: Modular solidity library contracts for flexibility


RayHuang880301:
**Is your feature request related to a problem? Please describe.**
As a smart contract developer who uses the library contract from `zk-kit`, I hope I can choose to import this library as an internal function to my main contract or as a publicly deployed library. It may be used in different ways based on the requirements (like I want to save the deployment gas cost or run time gas cost...)

**Describe the solution you'd like**
I propose to modular the library contract by changing it to consist of public and internal libraries. The core logic is implemented in the internal function of the internal library and the public library contract just calls to the internal library.

For example, the `IncrementalBinaryTree.sol` library can be moved to an internal function library such as `IncrementalBinaryTreeInternal.sol`, and all the original implementation remains the same, just changing its naming and function visibility from `public` to `internal`
Then use the public `IncrementalBinaryTree.sol` to inherit `IncrementalBinaryTreeInternal.sol`, which means that the actual implementation is in `IncrementalBinaryTreeInternal.sol` and `IncrementalBinaryTree.sol` is only a public function to call the implementation in `IncrementalBinaryTreeInternal.sol`

 In practice, it is roughly as follows
<img width="697" alt="image" src="https://github.com/privacy-scaling-explorations/zk-kit/assets/84802160/c8eef336-1595-4c41-a556-a167393463ac">

There are some benefits to doing this
1. For developers to be more flexible, they can choose to use external libraries or internal libraries
2. For developers who want to use internal libraries to save run time gas costs, they can directly inherit `IncrementalBinaryTreeInternal.sol` to their main contract
3. Those who want to save deployment costs can directly use the deployed public shared `IncrementalBinaryTree` library which is provided the PSE website or github

In addition to the above, this approach maybe can be applied to many contracts in `zk-kit` generally, like `IncrementalQuinTree.sol`, `LazyMerkleTree.sol`... etc. 😉

0xjei:
> **Is your feature request related to a problem? Please describe.** As a smart contract developer who uses the library contract from `zk-kit`, I hope I can choose to import this library as an internal function to my main contract or as a publicly deployed library. It may be used in different ways based on the requirements (like I want to save the deployment gas cost or run time gas cost...)
> 
> **Describe the solution you'd like** I propose to modular the library contract by changing it to consist of public and internal libraries. The core logic is implemented in the internal function of the internal library and the public library contract just calls to the internal library.
> 
[...]

PR #92, opened 2023-12-04
Modularize `contracts` package

0xjei:
## Description

This PR is intended to bring about the necessary improvements as stated in proposal #66.

To Do
- Create a `Constants.sol` file for the SNARK_SCALAR_FIELD and MAX_DEPTH constants.
- [ ] Update IMT implementations according to the proposal.
- [ ] Update tests
- [ ] Run benchmarks 

## Related Issue

See #66

-   [ ] Yes
-   [x] No

## Other information
None


0xjei:
### Before
Screenshot from 2023-12-04 11-57-56

### After
image

cedoor:
Hey @RayHuang880301 Could you also take a look at this PR to see if it's what you expected?

RayHuang880301:
Hey @0xjei @cedoor 
Thank you very much for your work. It seems that this is exactly what I expected. 👍

> Libraries, from what I know, cannot be inherited. So, how could you use an internal function of a library in a contract?

Because a `library` is used here instead of a `contract`, so as you said, `import` should be used here instead of `inherit`. It looks good to me now. Thank you!

0xjei:
> Hey @0xjei @cedoor Thank you very much for your work. It seems that this is exactly what I expected. 👍
> 
> > Libraries, from what I know, cannot be inherited. So, how could you use an internal function of a library in a contract?
> 
> Because a `library` is used here instead of a `contract`, so as you said, `import` should be used here instead of `inherit`. It looks good to me now. Thank you!

Thank you @RayHuang880301 :) @cedoor, green lights

cedoor (approved):

PR #111, opened 2023-12-22
feat(poseidon-cipher): implement poseidon encryption and decryption

### LINKED ISSUE #110: Poseidon encryption and decryption
**Describe the package you'd like**

Implement a TypeScript package which can perform encryption and decryption following this [paper]( https://drive.google.com/file/d/1EVrP3DzoGbmzkRmYnyEDcIQcXVU7GlOd/view).

**Additional context**

[Original implementation](https://github.com/weijiekoh/circomlib/tree/feat/poseidon-encryption/src)

ctrlc03:
## Description

This new package implements an encryption and a decryption function using the Poseidon hash permutations. It follows this [paper](https://drive.google.com/file/d/1EVrP3DzoGbmzkRmYnyEDcIQcXVU7GlOd/view) and is a TS port of this [code](https://github.com/weijiekoh/circomlib/tree/feat/poseidon-encryption/src)

## Related Issue

fix #110 

-   [ ] Yes
-   [x] No


packages/poseidon-cipher/README.md
```diff
@@ -0,0 +1,89 @@
...
+<script src="https://cdn.jsdelivr.net/npm/@zk-kit/poseidon-cipher"></script>
+```
+
+## Contacts
+
+### Developers
+
+-   e-mail : me@cedoor.dev
+-   github : [@cedoor](https://github.com/cedoor)
+-   website : https://cedoor.dev
```
cedoor, line 89: This can be removed. Or you can add yourself if you want @ctrlc03

packages/poseidon-cipher/package.json
```diff
@@ -0,0 +1,57 @@
...
+        "README.md",
+        "LICENSE"
+    ],
+    "repository": "git@github.com:privacy-scaling-explorations/zk-kit.git",
+    "homepage": "https://github.com/privacy-scaling-explorations/zk-kit/tree/main/packages/poseidon-cipher",
+    "author": {
+        "name": "Cedoor",
+        "email": "me@cedoor.dev",
+        "url": "https://cedoor.dev"
+    },
```
cedoor, line 37: Same as above.

cedoor (changes requested):
Nice package <3

ctrlc03:
> Nice package <3

🙏🙏
Would you mind taking a look at why it doesn't compile when you have some time please? It complains about ffjavascript not exporting utils, though it's the same version used in other packages and it's used in the same way 😔

cedoor:
> 🙏🙏 Would you mind taking a look at why it doesn't compile when you have some time please? It complains about ffjavascript not exporting utils, though it's the same version used in other packages and it's used in the same way 😔

It's only using `unstringifyBigInts` right? What do you think about re-implementing it in the `utils.ts` file?

ctrlc03:
> > 🙏🙏 Would you mind taking a look at why it doesn't compile when you have some time please? It complains about ffjavascript not exporting utils, though it's the same version used in other packages and it's used in the same way 😔
> 
> It's only using `unstringifyBigInts` right? What do you think about re-implementing it in the `utils.ts` file?

good shout, updated!

packages/poseidon-cipher/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "scripts": {
+        "build:watch": "rollup -c rollup.config.ts -w --configPlugin typescript",
+        "build": "rimraf dist && rollup -c rollup.config.ts --configPlugin typescript",
+        "prepublishOnly": "yarn build"
+    },
+    "publishConfig": {
+        "access": "public"
+    },
+    "devDependencies": {
+        "@zk-kit/eddsa-poseidon": "workspace:^",
```
cedoor, line 42: I'm not 100% sure that NPM and Yarn resolve this version when the package is published. We can try tho.

cedoor:
> good shout, updated! ps. the issue was with having removed `author` from package.json, and the rollup config script erroring out when generating the banner

Ok great! I can still see some Eslint errors tho

packages/poseidon-cipher/package.json
ctrlc03: should I put the exact version?

cedoor: I think it's worth trying it. If it works we could adopt the same approach with the other packages too.

ctrlc03: Let's see 👀 eslint and everything else should be fixed now

packages/poseidon-cipher/src/data/poseidonConstants.json
cedoor, line 1: @ctrlc03 if you want this package to also be used on browsers this file could be converted to ts, similarly to these: https://github.com/vimwitch/poseidon-lite/blob/main/src/constants/1.mjs (`poseidon-lite` formatted them as base64 strings to reduce their size I think). So that you can import it directly and it doesn't need to depend on `fs` and `path` anymore.

packages/poseidon-cipher/rollup.config.ts
```diff
@@ -0,0 +1,39 @@
...
+            name,
+            format: "iife",
+            banner
+        },
+        {
+            file: pkg.unpkg,
+            name,
+            format: "iife",
+            plugins: [terser({ output: { preamble: banner } })]
+        },
```
cedoor, line 31: These are not necessary here as there's another rollup config file for `iife`.

cedoor (changes requested):

packages/poseidon-cipher/src/data/poseidonConstants.json
ctrlc03: that's neat - I converted to ts for now, though feel like keeping all in one file might be a bit cleaner for this package, what you think?

cedoor (approved):

packages/poseidon-cipher/src/data/poseidonConstants.json
cedoor: Yes, I agree!

packages/poseidon-cipher/package.json
```diff
@@ -0,0 +1,51 @@
...
+        "dist/",
+        "src/",
+        "README.md",
+        "LICENSE"
+    ],
+    "repository": "git@github.com:privacy-scaling-explorations/zk-kit.git",
+    "homepage": "https://github.com/privacy-scaling-explorations/zk-kit/tree/main/packages/poseidon-cipher",
+    "scripts": {
+        "build:watch": "rollup -c rollup.config.ts -w --configPlugin typescript",
+        "build": "rimraf dist && rollup -c rollup.config.ts --configPlugin typescript",
```
cedoor, line 35: This needs to be updated. It should be like: https://github.com/privacy-scaling-explorations/zk-kit/blob/a2863c71dfdc5ac5790b80760c9d601c484f6140/packages/baby-jubjub/package.json#L28

cedoor (changes requested):

packages/poseidon-cipher/package.json
ctrlc03: good catch, thank you

cedoor (approved):

PR #115, opened 2024-01-04
Packed points with less than 32 bytes are now padded

cedoor:
## Description

When converting a packed point with less than 32 bytes from bigint to hex or buffer it is necessary to pad it so that the number of bytes is always 32. 

-   [ ] Yes
-   [x] No


ctrlc03 (approved):
🙏 LGTM

PR #124, opened 2024-01-12
Add `NUMBER_OF_INPUTS` to `poseidon-proof`

### LINKED ISSUE #116: Add `numberOfInputs` parameter to PoseidonProof circuit
The `poseidon-proof` circuit can have a parameter that specify the number of inputs the Poseidon hash function can have. The nullified parameters could be changed so that the second parameter is not the preimage (or the preimages in case of many inputs), but the digest, since the digest is a public output.

0xjei:
## Related Issue
Closes #116

-   [ ] Yes
-   [x] No

## Other Information
This PR introduces the use of `digest` to compute the `nullifier` output signal instead of the `preimage` as specified in the #116.

Note that the number of constraints for `NUMBER_OF_INPUTS = 1` is the same as in the previous implementation. As this number increases, the complexity of the circuit increases.

The `@zk-kit/poseidon-proof` package will be updated accordingly.

__Todos__
- [x] Update the 'getArtifacts' methods based on the number
of inputs.
- [x] Update the 'circomkit' configuration to support 'scope'
as a public signal.
- [x] Bug: In the 'poseidon-proof' package, 'digest' and 'nullifier' are inverted in the return of the 'generate' method.
- [x] Generate artifacts for a different number of inputs (1 <= noi
<= 16).
- [x] Make 'verification-keys.json' support the corresponding key per input (matrix for 'vk_delta_2' and
'IC').
- [x] Add missing tests.
- [x] Update remote artifacts @cedoor


packages/circuits/circom/poseidon-proof.circom
```diff
@@ -5,20 +5,24 @@ include "poseidon.circom";
 // This circuit can be used to prove the possession of a pre-image of a
```
cedoor, line 5:
```suggestion
// This circuit can be used to prove the possession of pre-images of a
```

packages/circuits/circom/poseidon-proof.circom
```diff
@@ -5,20 +5,24 @@ include "poseidon.circom";
...
 
+    assert (NUMBER_OF_INPUTS >= 1);
+
     // It applies the Poseidon hash function to the pre-image to produce a hash digest.
     signal output digest;
-    digest <== Poseidon(1)([preimage]);
+    digest <== Poseidon(NUMBER_OF_INPUTS)(preimages);
 
-    // A nullifier is also computed using both the pre-image and the scope, providing a value
+    // A nullifier is also computed using both the pre-image and the digest, providing a value
```
cedoor, line 24:
```suggestion    
    // A nullifier is also computed using both the scope and the digest, providing a value
```

cedoor (changes requested):
Great job, just a few corrections in the comments!

0xjei:
Converted to draft to make changes to `@zk-kit/poseidon-proof` before merging this feature.

0xjei:
Ready for review @cedoor

cedoor (changes requested):
Great job @0xjei! Just a small fix for a bug that was already there before your PR.

cedoor (approved):

PR #126, opened 2024-01-13
fix(imt.sol): add correct indices in insert many function

### LINKED ISSUE #125: Wrong indices when inserting many members
## Issue

The `insertMany` function uses the real position of the leaves in an array, this should be updated to use the real position + 1.

## Context

The leaves indices in the LeanIMT contracts are not the same as the ones returned with the function `indexOf`. The indices start at position 1 instead of 0 to facilitate checks for leaf existence and retrieval of leaf position. Then the first leaf is in position 0, but the saved index is 1 and when the `indexOf` function is used with this leaf, it will return 0.

vplasencia:
## Related Issue

Closes #125 

-   [ ] Yes
-   [x] No

### DIFF
--- a/packages/imt.sol/contracts/internal/InternalLeanIMT.sol
+++ b/packages/imt.sol/contracts/internal/InternalLeanIMT.sol
@@ -90,7 +90,7 @@ library InternalLeanIMT {
                 revert LeafAlreadyExists();
             }
 
-            self.leaves[leaves[i]] = self.size + i;
+            self.leaves[leaves[i]] = self.size + 1 + i;
 
             unchecked {
                 ++i;
--- a/packages/imt.sol/test/LeanIMT.ts
+++ b/packages/imt.sol/test/LeanIMT.ts
@@ -280,6 +280,15 @@ describe("LeanIMT", () => {
 
             expect(index).to.equal(0)
         })
+        it("Should return the indices of the leaves", async () => {
+            await leanIMTTest.insertMany([1, 2])
+
+            const index1 = await leanIMTTest.indexOf(1)
+            const index2 = await leanIMTTest.indexOf(2)
+
+            expect(index1).to.equal(0)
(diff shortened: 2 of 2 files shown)


packages/imt.sol/contracts/internal/InternalLeanIMT.sol
```diff
@@ -90,7 +90,7 @@ library InternalLeanIMT {
                 revert LeafAlreadyExists();
             }
 
-            self.leaves[leaves[i]] = self.size + i;
```
0xjei, line 93: Can you explain why adding '+ 1' is correct here?

I couldn't find a concrete answer, probably I'm missing some implementation details. Following your test below, I expected `indexOf(1) = 1` due to `self.size = 0` and `i = 0`, but it seems to return `indexOf(1) = 0` as you wrote. Hmm, just waiting for your feedback and making this a general comment.

vplasencia: Hey! You can find more info in the issue description #125.

The function `indexOf` should work like the JavaScript `indexOf` function. In JavaScript, if we have the array `[1, 2]`, `indexOf(1)` should return 0.

The values saved before were the real positions (`self.leaves[leaves[i]] = self.size + i;`), but the positions that should be saved in `self.leaves` is the real one + 1 (`self.leaves[leaves[i]] = self.size + 1 + i;`). That's why the `indexOf` function in the contracts returns `self.leaves[leaf] - 1`.

Let me know if it is more clear now.

cedoor (approved):

packages/imt.sol/contracts/internal/InternalLeanIMT.sol
0xjei: Thank you so much for the explanation. Now makes definitely much more sense :blush:

0xjei (approved):

PR #127, opened 2024-01-15
Add eddsa-proof package

vplasencia:
## Related Issue

#105 

-   [ ] Yes
-   [x] No


0xjei:
scheduled review on Thursday

packages/circuits/circom/eddsa-proof.circom
```diff
@@ -0,0 +1,31 @@
...
+
+    // Get the two Baby Jubjub points using the private key.
+    (Ax, Ay) = BabyPbk()(secret);
+
+    // It applies the Poseidon hash function to the to Baby Jubjub poits to produce the commitment.
+    commitment <== Poseidon(2)([Ax, Ay]);
+
+    // A nullifier is also computed using both the scope and the commitment, providing a value
+    // to prevent the same proof from being reused twice.
+    nullifier <== Poseidon(2)([scope, commitment]);
```
cedoor, line 30: @vplasencia @0xjei 

I was thinking a bit more about this nullifier that uses public pre-images. I think it might not need it at all. Externally one could just use the scope and commitment (which are public) and hash them to obtain the same nullifiers. Why should that value be inside the circuit? I don't see any advantages.

Let me know if it makes sense or if I'm missing something.

packages/eddsa-proof/README.md
```diff
@@ -0,0 +1,149 @@
...
+// Eventually you may want to check the nullifier.
+```
+
+## 📈 Benchmarks
+
+Benchmarks were run on a MacBook Pro, Apple M2 Pro, 16 GB RAM machine, after initializing the BN128 curve with [`@zk-kit/groth16`](https://github.com/privacy-scaling-explorations/zk-kit/edit/main/packages/groth16)-`buildBn128` (~230ms).
+
+| Generate proof | Verify proof | Constraints |
+| -------------- | ------------ | ----------- |
+| `167.652 ms`   | `10.337 ms`  | `1256`      |
```
cedoor, line 130: Nice!

packages/eddsa-proof/eddsa-proofVerifier.sol
cedoor, line 1: What about naming this file just "Verifier.sol"?

packages/eddsa-proof/package.json
```diff
@@ -0,0 +1,59 @@
+{
+    "name": "@zk-kit/eddsa-proof",
+    "version": "0.5.0",
```
cedoor, line 3:
```suggestion
    "version": "0.1.0",
```

packages/eddsa-proof/package.json
```diff
@@ -0,0 +1,59 @@
...
+        "dist/",
+        "src/",
+        "LICENSE",
+        "README.md"
+    ],
+    "engines": {
+        "node": ">=18.0.0"
+    },
+    "repository": "git@github.com:privacy-scaling-explorations/zk-kit.git",
+    "homepage": "https://github.com/privacy-scaling-explorations/zk-kit/tree/main/packages/poseidon-proof",
```
cedoor, line 27:
```suggestion
    "homepage": "https://github.com/privacy-scaling-explorations/zk-kit/tree/main/packages/eddsa-proof",
```

packages/eddsa-proof/rollup.browser.config.ts
```diff
@@ -0,0 +1,38 @@
...
+import * as fs from "fs"
+import cleanup from "rollup-plugin-cleanup"
+import typescript from "rollup-plugin-typescript2"
+
+const pkg = JSON.parse(fs.readFileSync("./package.json", "utf-8"))
+const banner = `/**
+ * @module ${pkg.name}
+ * @version ${pkg.version}
+ * @file ${pkg.description}
+ * @copyright Ethereum Foundation 2023
```
cedoor, line 12:
```suggestion
 * @copyright Ethereum Foundation 2024
```

We could use `new Date().getFullYear()` in all the packages.

packages/eddsa-proof/rollup.node.config.ts
```diff
@@ -0,0 +1,40 @@
...
+import * as fs from "fs"
+import cleanup from "rollup-plugin-cleanup"
+import typescript from "rollup-plugin-typescript2"
+
+const pkg = JSON.parse(fs.readFileSync("./package.json", "utf-8"))
+const banner = `/**
+ * @module ${pkg.name}
+ * @version ${pkg.version}
+ * @file ${pkg.description}
+ * @copyright Ethereum Foundation 2023
```
cedoor, line 11:
```suggestion
 * @copyright Ethereum Foundation 2024
```

packages/eddsa-proof/src/generate.ts
```diff
@@ -0,0 +1,53 @@
...
+ * of a Semaphore identity of a group of voters, the scope could be the poll's ID.
+ * @param privateKey The private key of the commitment.
+ * @param scope A public value used to contextualize the cryptographic proof
+ * and calculate the nullifier.
+ * @param snarkArtifacts The Snark artifacts (wasm and zkey files) generated in
+ * a trusted setup of the circuit are necessary to generate valid proofs
+ * @returns The Poseidon zero-knowledge proof.
+ */
+export default async function generate(
+    privateKey: string | number | bigint,
```
cedoor, line 26: Should we support the same types as the `@zk-kit/eddsa-poseidon` package?

https://github.com/privacy-scaling-explorations/zk-kit/blob/a30fa3b65f3508df6a84fd6b8444f98a054e5977/packages/eddsa-poseidon/src/eddsa-poseidon.ts#L37

It also supports `Buffer`s.

packages/eddsa-proof/src/get-snark-artifacts.node.ts
```diff
@@ -0,0 +1,22 @@
...
+import download from "download"
+import tmp from "tmp"
+import fs from "fs"
+import { defaultSnarkArtifacts } from "./config"
+
+export default async function getSnarkArtifacts() {
+    const tmpDir = "eddsa-proof"
+    const tmpPath = `${tmp.tmpdir}/${tmpDir}`
+
+    if (!fs.existsSync(tmpPath)) {
```
cedoor, line 11: It misses some checks that were recently added here:

https://github.com/privacy-scaling-explorations/zk-kit/blob/a30fa3b65f3508df6a84fd6b8444f98a054e5977/packages/poseidon-proof/src/get-snark-artifacts.node.ts#L11-L24

cedoor (changes requested):
I've added a few comments and suggestions. Overall, LGTM! Great work 🔥

packages/eddsa-proof/src/get-snark-artifacts.node.ts
vplasencia: 1- Maybe this part is not necessary because the `download` package should create the folder:
https://github.com/privacy-scaling-explorations/zk-kit/blob/a30fa3b65f3508df6a84fd6b8444f98a054e5977/packages/poseidon-proof/src/get-snark-artifacts.node.ts#L11-L13

2- Regarding the other check, the issue I see is that if someone updates the package in an existing project and the artifacts are different, then the project will not get the new artifacts:
https://github.com/privacy-scaling-explorations/zk-kit/blob/a30fa3b65f3508df6a84fd6b8444f98a054e5977/packages/poseidon-proof/src/get-snark-artifacts.node.ts#L15

Let me know what you think.

packages/circuits/circom/eddsa-proof.circom
vplasencia: Hey yes, true. Since the `scope` and `commitment` are public signals, it is not necessary to generate the nullifier there, since it can be generated outside the circuit using the public signals.

packages/eddsa-proof/src/get-snark-artifacts.node.ts
cedoor: 1. Umh could you check?

2. That can only happen if a dev who is working on zk-kit updates the circuit right? I think it'll happen rarely, and in those cases, they could remove that folder manually. Otherwise, the only solution for this would be overwriting those artifacts every time a proof is generated, which makes it much less fast.

vplasencia: 1- I double checked and yes, the folder is created if it does not exist.

2- Yes, exactly, that will only happen if the artifacts change. I think we can add that check and keep this issue in mind.

cedoor: Oh ok 👍🏽

vplasencia: We only needed to check the existence of the folder because `download` can still create the folder, then we only need this: 

```ts
if (!fs.existsSync(tmpPath) || fs.readdirSync(tmpPath).length !== 2) {
```

cedoor (approved):

packages/circuits/circom/eddsa-proof.circom
cedoor: Great, it looks like we can remove them from the `poseidon-proof` circuit too then!

packages/circuits/circom/eddsa-proof.circom
```diff
@@ -0,0 +1,26 @@
...
+// This circuit can be used to prove the possession of a private key of an
+// identity commitment without revealing the private key itself. It utilizes 
+// the {@link https://eips.ethereum.org/EIPS/eip-2494|Baby Jubjub} elliptic curve, the Poseidon hash function which a highly efficient 
+// and secure hash function suited for zero-knowledge proof contexts.
+// A scope value can be used to define a nullifier to prevent the same
+// proof from being re-used twice.
+template EddsaProof() {
+    // The circuit takes two inputs: the private key and an additional scope parameter.
+    signal input secret;
+    signal input scope;
```
0xjei, line 15: It appears that the `scope` is not being used in the circuit. If this is the case, it can be removed. @cedoor @vplasencia 

If that's the case, we should adapt the tests accordingly and `eddsa-proof` package as well.

0xjei (changes requested):
Awesome @vplasencia :fire: 

Please, review the `scope` signal usage - everything else LGTM! :+1:

packages/circuits/circom/eddsa-proof.circom
cedoor: I think it should stay there to allow devs to use it to prevent double-signaling. As we've discussed, nullifier = scope + commitment.

It needs a dummy square, like the `message` signal in the Semaphore circuit -> https://github.com/semaphore-protocol/semaphore/blob/596e10c210e5e94cb2b2d79089d1add92d52d1a2/packages/circuits/semaphore.circom#L24

0xjei: Makes sense and thank you for the explanation. Yes, we can just add the dummy square and that's it.

cedoor: Yess, I've opened another issue for `poseidon-proof`: https://github.com/privacy-scaling-explorations/zk-kit/issues/136

0xjei (approved):

cedoor (approved):

PR #130, opened 2024-01-20
feat(ecdh): implement ECDH circom circuit

ctrlc03:
## Description

Implement a circom circuit that allows to generate a shared secret from a private key and a public key.

-   [ ] Yes
-   [x] No

### DIFF
--- a/packages/circuits/circom/circuits.json
+++ b/packages/circuits/circom/circuits.json
@@ -29,5 +29,11 @@
         "file": "poseidon-cipher",
         "template": "PoseidonPerm",
         "params": [2]
+    },
+    "ecdh": {
+        "file": "ecdh",
+        "template": "Ecdh",
+        "pubs": ["publicKey"],
+        "params": []
     }
 }
--- a/packages/circuits/circom/ecdh.circom
+++ b/packages/circuits/circom/ecdh.circom
@@ -1,0 +1,30 @@
+pragma circom 2.1.5;
+
+// circomlib imports
+include "./bitify.circom";
+include "./escalarmulany.circom";
+
+// ECDH Is a a template which allows to generate a shared secret
+// from a private key and a public key
+// on the baby jubjub curve
+// It is important that the private key is hashed and pruned first
+// which can be accomplished using the function 
+// deriveScalar from @zk-kit/baby-jubjub
+template Ecdh() {
+    // the private key must pass through deriveScalar first 
+    signal input privateKey;
+    signal input publicKey[2];
+    
+    signal output sharedKey[2];
+    
+    // convert the private key to its bits representation
+    var out[253];
+    out = Num2Bits(253)(privateKey);
+
+    // multiply the public key by the private key
+    var mulFix[2];
+    mulFix = EscalarMulAny(253)(out, publicKey);
+
+    // we can then wire the output to the shared secret signal
+    sharedKey <== mulFix;
+}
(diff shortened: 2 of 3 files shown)


0xjei:
scheduled review on Thursday

packages/circuits/tests/ecdh.test.ts
```diff
@@ -0,0 +1,82 @@
...
+    let circuit: WitnessTester<["privateKey", "publicKey"], ["sharedKey"]>
+
+    before(async () => {
+        circuit = await circomkit.WitnessTester("ecdh", {
+            file: "ecdh",
+            template: "Ecdh"
+        })
+    })
+
+    it("should correctly compute a ECDH shared key", async () => {
```
0xjei, line 16:
```suggestion
    it("should correctly compute an ECDH shared key", async () => {
```

packages/circuits/tests/ecdh.test.ts
```diff
@@ -0,0 +1,82 @@
...
+
+        const circuitInputs = {
+            privateKey: BigInt(deriveSecretScalar(privateKey1)),
+            publicKey: publicKey2
+        }
+
+        await circuit.expectPass(circuitInputs, { sharedKey: [ecdhSharedKey[0], ecdhSharedKey[1]] })
+    })
+
+    it("should geneate the same shared key from the same keypairs", async () => {
```
0xjei, line 32:
```suggestion
    it("should generate the same shared key from the same keypairs", async () => {
```

packages/circuits/circom/ecdh.circom
```diff
@@ -0,0 +1,38 @@
...
+
+    // need to loop for the length of the private key
+    for (var i = 0; i < 253; i++) {
+        mulFix.e[i] <== privBits.out[i];
+    }
+
+    // we can then wire the output to the shared secret signal
+    sharedKey[0] <== mulFix.out[0];
+    sharedKey[1] <== mulFix.out[1];
+}
```
0xjei, line 37: We can use [Anonymous Components](https://docs.circom.io/circom-language/anonymous-components-and-tuples/) here.

```suggestion
template Ecdh() {
    // the private key must pass through deriveScalar first 
    signal input privateKey;
    signal input publicKey[2];
    
    signal output sharedKey[2];
    
    // convert the private key to its bits representation
    var out[253];
    out = Num2Bits(253)(privateKey);

    // multiply the public key by the private key
    var mulFix[2];
    mulFix = EscalarMulAny(253)(out, publicKey);

    // we can then wire the output to the shared secret signal
    sharedKey <== mulFix;
}
```

packages/circuits/circom/circuits.json
```diff
@@ -29,5 +29,10 @@
         "file": "poseidon-cipher",
         "template": "PoseidonPerm",
         "params": [2]
+    },
+    "ecdh": {
+        "file": "ecdh",
+        "template": "Ecdh",
+        "params": []
```
0xjei, line 36: Is `publicKey` a public signal? If yes, we should add it to the `pubs`.

```suggestion
        "file": "ecdh",
        "template": "Ecdh",
        "pubs": ["publicKey"],
        "params": []
```

0xjei (changes requested):
Thank you @ctrlc03 :fire: 

Left some comments and suggestions! Let me know :blush:

packages/circuits/circom/ecdh.circom
ctrlc03: oh so that's how they work 😆 great stuff thank you

packages/circuits/circom/circuits.json
ctrlc03: I'm not sure here, it's never been used on its own. But it does make sense for it to be a public signal

cedoor (approved):
Hey, I'd just recommend using the anonymous components like @0xjei already did!

In general, LGTM 👍🏽

0xjei (approved):

PR #134, opened 2024-01-29
feat(imt): add rust implementation of imt

### LINKED ISSUE #133: Build a Rust implementation of the IMTs

Arch0125:
## Description

This PR adds Rust implmentation of IMT, aligned with the current TS version, with the following methods included insert/update/delete/create-proof/verify-proof

## Related Issue

Resolves #133 

-   [ ] Yes
-   [x] No

### Test coverage
- [x] **Test New IMT Instance**
  - Description: Ensure that a new `IMT` instance is created successfully.

- [x] **Test Leaf Insertion**
  - Description: Check if new leaves can be inserted into the tree.

- [x] **Test Leaf Deletion**
  - Description: Test the deletion functionality of a leaf in the tree.

- [x] **Test Leaf Update**
  - Description: Ensure that a leaf in the tree can be updated correctly.

- [x] **Test Create and Verify Proof**
  - Description: Create a Merkle proof for a leaf and verify its correctness.

- [x] **Should Not Initialize with Too Many Leaves**
  - Description: The tree should not initialize if the number of leaves exceeds `arity ^ depth`.

- [x] **Should Not Insert in Full Tree**
  - Description: Test that leaves cannot be inserted into a full tree.

- [x] **Should Not Delete Nonexistent Leaf**
  - Description: Ensure that attempting to delete a nonexistent leaf results in an error.


packages/imt-rs/src/lib.rs
```diff
@@ -0,0 +1,320 @@
...
+
+    pub fn leaves(&mut self) -> Vec<Option<String>> {
+        return self.nodes[0].clone();
+    }
+
+    pub fn arity(&mut self) -> usize {
+        return self.arity;
+    }
+
+    pub fn insert(&mut self, leaf: IMTNode) -> Result<(), &'static str> {
```
curryrasul, line 80: I think would be good to try to change insert function with insertion to vec, and then calling the `update` function to update/recalculate the tree

packages/imt-rs/src/lib.rs
```diff
@@ -0,0 +1,320 @@
...
+    root: IMTNode,
+    leaf: IMTNode,
+    path_indices: Vec<usize>,
+    siblings: Vec<Vec<IMTNode>>,
+    leaf_index: usize,
+}
+
+
+impl IMT {
+    pub fn new(hash: IMTHashFunction, depth: usize, zero_value: IMTNode, arity: usize, leaves: Option<Vec<IMTNode>>) -> Result<IMT, &'static str> {
```
curryrasul, line 23: we can take `Vec<IMTNode>>` instead of `Option<Vec<_>>`, cause we simply can pass empty vec, if we don't have any leaves to insert during initialization

packages/imt-rs/src/lib.rs
```diff
@@ -0,0 +1,320 @@
...
+                }
+            }
+        }
+
+        Ok(imt)
+    }
+
+    pub fn root(&mut self) -> Option<String>  {
+        let nodes = self.nodes[self.depth].clone();
+        return nodes[0].clone();
```
curryrasul, line 65: `return` usually not used in the end of the function; this applies to other functions as well

packages/imt-rs/src/lib.rs
curryrasul, line 1: I'd suggest minor changes in the current implementation (I showed few directly). For that I'd recommend using Clippy tool. It's basically a linter, that shows how you can fix/change stuff to follow Rust convention of code. You should also use cargo-fmt tool, if you don't do that.

But overall, Is this implementation follows TS implementation fully? I mean not for API but for how is everything implemented inside. 
Cause I'd store the tree not as a Vector of Vectors, but for example as a HashMap, where key is a tuple `(depth, position)` and value is a following leaf/node. Also we don't need to really **initialize** all tree in the new function, etc. I'd suggest to check how it's done in [pmtree](https://github.com/Rate-Limiting-Nullifier/pmtree/blob/main/src/tree.rs) .
If you don't want to change the core structure - then everything's good, except minor fixes I mentioned earlier. 
Thanks for attention!

curryrasul (changes requested):

cedoor:
Hey @Arch0125, do you think you'll have time to work on the remaining minor changes?

We're planning to create a ZK-Kit repo for Rust and it would be nice to have this implementation there :)

Arch0125:
Hey @cedoor, yeah ill get done with the minor changes mostly by EOD, or by earlier tomorrow.
It kinda slipped out of my mind, got busy with some stuff at work, apologies for that

cedoor:
@Arch0125 no worries :) Thanks for your quick response!

packages/imt-rs/src/lib.rs
0xjei: looking at the `pmtree` implementation you mentioned here, do you think this is still valuable as `zk-kit.rust` package? @curryrasul

packages/imt-rs/src/lib.rs
```diff
@@ -0,0 +1,320 @@
+use std::error::Error;
```
0xjei, line 1: we can get rid of this since it's not used.

packages/imt-rs/src/lib.rs
```diff
@@ -0,0 +1,320 @@
...
+
+pub struct IMT {
+    nodes: Vec<Vec<Option<IMTNode>>>,
+    zeroes: Vec<IMTNode>,
+    hash: IMTHashFunction,
+    depth: usize,
+    arity: usize,
+}
+
+struct IMTMerkleProof {
```
0xjei, line 13: should we mark this as `public` too? seems treated as private type here but thought as public. not really sure, forwarding to @curryrasul

packages/imt-rs/src/lib.rs
```diff
@@ -0,0 +1,320 @@
...
+            for i in level_start_index..level_end_index {
+                if i < self.nodes[level].len() {
+                    children.push(self.nodes[level][i].clone().unwrap_or_else(|| self.zeroes[level].clone()));
+                } else {
+                    children.push(self.zeroes[level].clone());
+                }
+            }
+
+            node = (self.hash)(children);
+            index = index / self.arity;
```
0xjei, line 114:
```suggestion
            index /= self.arity;
```

packages/imt-rs/src/lib.rs
```diff
@@ -0,0 +1,320 @@
...
```
0xjei, line 163:
```suggestion
            index /= self.arity;
```

packages/imt-rs/src/lib.rs
cedoor: @0xjei, chatted about that with @curryrasul. We're going to add it to the zk-kit.rust repo as the first package :)

packages/imt-rs/src/lib.rs
```diff
@@ -0,0 +1,320 @@
...
+    leaf_index: usize,
+}
+
+
+impl IMT {
+    pub fn new(hash: IMTHashFunction, depth: usize, zero_value: IMTNode, arity: usize, leaves: Option<Vec<IMTNode>>) -> Result<IMT, &'static str> {
+
+        if leaves.is_some() && leaves.as_ref().unwrap().len() > arity.pow(depth as u32) {
+            return Err("The tree cannot contain more than arity^depth leaves");
+        }
```
0xjei, line 27: you can get advantage of the proper [pattern matching](https://rust-book.cs.brown.edu/ch18-03-pattern-syntax.html) here
```suggestion
        if let Some(ref leaves) = leaves {
            if leaves.len() > arity.pow(depth as u32) {
                return Err("The tree cannot contain more than arity^depth leaves");
            }
        }
```

0xjei (changes requested):
hey @Arch0125 tysm for your work here. I avoided some comments as they were redundant from @curryrasul's review.

In any case, regardless of the design of the solution, I see that the `lib.rs` is a bit too loaded. It would be nice to have something like:

```rust
// this is lib.rs.

pub mod imt;
pub mod hash;
```

by moving the necessary code into `imt.rs` and `hash.rs` respectively.

would love to hear your thoughts here as well @curryrasul

packages/imt-rs/src/lib.rs
0xjei: sweet, tysm for the reply @cedoor

Arch0125:
Updating the code structure in the next commit as mentioned by @0xjei

0xjei (approved):
lgtm, tysm @Arch0125

cedoor:
@Arch0125 This PR should be ready right? Could you solve the conflicts?

Arch0125:
Yep solving the conflicts asap

Arch0125:
@cedoor ive fixed the remaining issues that came up after the merge

cedoor:
@Arch0125 Thank you so much for this PR! As a second step, we need to move this crate to the [zk-kit.rust](https://github.com/privacy-scaling-explorations/zk-kit.rust) repo. Would you like to do it yourself?

Arch0125:
Yeah sure, ill take it up, can you assign it there ?

PR #139, opened 2024-02-08
feat(poseidon-cipher): add decrypt without checks to not throw on failure

ctrlc03:
## Description

Implement copy of circuit.decryptWithoutCheck to replicate circuit logic in TS.

-   [ ] Yes
-   [x] No

### DIFF
--- a/packages/poseidon-cipher/package.json
+++ b/packages/poseidon-cipher/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@zk-kit/poseidon-cipher",
-    "version": "0.1.1",
+    "version": "0.2.0",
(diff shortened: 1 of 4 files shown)


packages/poseidon-cipher/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@zk-kit/poseidon-cipher",
-    "version": "0.1.1",
+    "version": "0.1.2",
```
cedoor, line 3: I'd suggest to bump a minor version here, as you're adding a new feature.

cedoor (approved):

PR #149, opened 2024-02-08
fix(utils): input mutations and small-number padding in BigInt <-> Buffer conversions

### LINKED ISSUE #147: leBufferToBigint reverses its input
**Describe the bug**

The buffer passed as argument to leBufferToBigint is reversed in-place, which could be a surprise for the caller.  This caused a lot of confusion when I was trying to test compatibility with circomlibjs by processing the same input in multiple ways.

**To Reproduce**
1. Make a buffer for input
2. Call `leBufferToBigint` on that input
3. Examine input buffer and see that its bytes have been reversed

I have a PR upcoming to fix this.  It'll include a unit test demonstrating this issue and the fix as soon as it's submitted.

**Expected behavior**
It's valid to argue that this is as-designed, since there's no documentation to demonstrate otherwise.  I argue that this is a potential source of bugs, and it's worth making an extra copy to avoid it.

**Additional context**
You can see some additional discussion on this issue between me and @cedoor on the #zk-kit channel of the PSE discord.
I work on Zupass at 0xPARC, and am in the process of evaluating zk-kit for use in some new PCDs.

### LINKED ISSUE #148: bigintToBuffer doesn't pad its output
**Describe the bug**

bigintToBuffer assumes its input is always full size, and writes its hex representation into the 32-byte output buffer at byte 0.  This produces incorrect results if the input bigint has a 0 in its most-significant byte.

**To Reproduce**
1. call bigintToBuffer(n) where n is a small number
2. reverse the conversion via BigInt(`0x${buffer.toString("hex")}`)
3. The output won't be your original number

I have a PR upcoming to fix this. It'll include a unit test demonstrating this issue and the fix as soon as it's submitted.

**Expected behavior**
bigintToBuffer should always produce a 32-byte buffer, and the input number should be right-aligned (big-endian conversion), padded with zeroes on the left.

**Additional context**
You can see some additional discussion on related issues (not this one, which came up during implementation) between me and @cedoor on the #zk-kit channel of the PSE discord.
I work on Zupass at 0xPARC, and am in the process of evaluating zk-kit for use in some new PCDs.

artwyman:
## Description

4 changes here, in order of importance:

1) While testing compatibility of zk-kit with circomlibjs, I discovered that `leBufferToBigint` has a side-effect of reversing its input.  This was surprising when I was trying to use the same input multiple times and compare results.  Fixing this introduces an extra copy of 32 bytes, but I think that's worthwhile to avoid the risk of bugs.  There are larger costs for the use of hex coversion via strings.

2) While testing this, I also noticed that `bigintToBuffer` doesn't always pad its results to 32 bytes, resulting in incorrect conversions.  I fixed this as well using an offset to the copy command.  These conversion functions seem to be intended for fixed-size numbers of 32 byes.  I'd recommend adding some more checks that input bigints aren't too large for this size, but I didn't try to tackle that in this PR.

3) I added `bufferToBigint` for symmetry, so utils can convert between Buffer and BigInt in both directions with either endinness.  This is mostly so my tests can be complete.

4) I added tests of the 4 conversion functions, including tests which demonstrate the bugs I fixed.  Note that I was told that `utils` tests are disabled in `jest.config.ts` because they are incomplete.  I enabled them locally so I could run my tests, then disabled them again when I was done, so that you can complete the rest of them on your own schedule.

## Related Issue

Fixes #147
Fixes #148

-   [ ] Yes
-   [x] No

## Other information

Note that the contribution guidelines say to submit a PR for the `dev` branch.  That doesn't seem to exist (or isn't visible to me) so I'm targeting `main`.

You can see some additional discussion on these issues (and others) between me and @cedoor on the #zk-kit channel of the PSE discord.

I work on Zupass at 0xPARC, and am in the process of evaluating zk-kit for use in some new PCDs.

### DIFF
--- a/packages/utils/package.json
+++ b/packages/utils/package.json
@@ -33,6 +33,7 @@
         "access": "public"
     },
     "devDependencies": {
+        "@types/chai": "^4.2.0",
         "rollup-plugin-cleanup": "^3.2.1",
         "rollup-plugin-polyfill-node": "^0.13.0",
         "rollup-plugin-terser": "^7.0.2",
--- a/packages/utils/src/number-conversions.ts
+++ b/packages/utils/src/number-conversions.ts
@@ -30,6 +30,18 @@ export function bufferToBigint(buffer: Buffer): bigint {
     return BigInt(`0x${buffer.toString("hex")}`)
 }
 
+export function bigintToBuffer(n: bigint): Buffer {
+    const hex = bigintToHexadecimal(n)
+
+    // Allocate buffer of the desired size, filled with zeros.
+    const buffer = Buffer.alloc(32, 0)
+
+    const fromHex = Buffer.from(hex, "hex")
+    fromHex.copy(buffer, 32 - fromHex.length)
+
+    return buffer
+}
+
 export function bigNumberishToBigint(value: BigNumberish): bigint {
     if (
         typeof value === "number" ||
@@ -44,7 +56,7 @@ export function bigNumberishToBigint(value: BigNumberish): bigint {
 }
 
 export function leBufferToBigint(buffer: Buffer): bigint {
-    return BigInt(`0x${buffer.reverse().toString("hex")}`)
+    return BigInt(`0x${Buffer.from(buffer).reverse().toString("hex")}`)
 }
 
 export function leBigintToBuffer(n: bigint): Buffer {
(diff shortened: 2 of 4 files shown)


cedoor:
Thanks a lot @artwyman 🙏🏽

cedoor (approved):
Looks great to me! Thanks for fixing these bugs 🙏🏽

Issue #123, opened 2024-01-11
Access Merkle Proof in LazyIMT


alrevuelta:
Question rather than feature, but it might lead to a feature if feasible.

I'm using the [LazyIMT](https://github.com/privacy-scaling-explorations/zk-kit/blob/96ceeba751756d8abaa179ad5c52afecf25c3fc7/packages/imt.sol/contracts/internal/InternalLazyIMT.sol) and was wondering if it would be possible to **get a Merkle Proof directly from the contract**. The main benefit I see, is that any app would be able to get said proof to prove things, without having to locally fetch the whole tree and construct it locally. I'm using the `LazyIMT` because it seems to offer the closest I'm looking for, but open for other trees.

With a reduced example, given the following tree with 8 leafs, I would like to get a Merkle Proof for a given leaf, **provided by the contract**. Example:
* Proof that `C` belongs to the tree: [`D`, `HASH_AB`, `HASH_EFGH`]
* Proof that `A` belongs to the tree: [`B`, `HASH_CD`, `HASH_EFGH`]

```
          ROOT (Hash_ABCDEFGH)
          /                  \
         /                    \
 HASH_ABCD                   HASH_EFGH
  /      \                   /       \
 /        \                 /         \
HASH_AB   HASH_CD      HASH_EF    HASH_GH
 /   \     /   \       /   \      /   \
A     B   C     D     E     F    G     H

```

The `LazyIMT` seems to store in [`elements`](https://github.com/privacy-scaling-explorations/zk-kit/blob/96ceeba751756d8abaa179ad5c52afecf25c3fc7/packages/imt.sol/contracts/internal/InternalLazyIMT.sol#L10) the whole tree, both leafs at the bottom and intermediate levels. In theory, I could get the Merkle Proof of `C`, by accessing `elements` at different levels:
  * `D`: `_indexForElement(0, 3)`
  * `HASH_AB`: `_indexForElement(1, 0)`
  * `HASH_EFGH`: `_indexForElement(2, 1)`

[...]

PR #162, opened 2024-02-12
feat: add get merkle proof to lazy imt tree

alrevuelta:
## Description

This PR adds a function to the `LazyIMT` contract, `_merkleProofElements`, which returns the Merkle proof of any leaf `index` in the tree. The `depth` is configurable, which allows to get the proofs of any subtree.

Closes https://github.com/privacy-scaling-explorations/zk-kit/issues/123

### DIFF
--- a/packages/imt.sol/contracts/internal/InternalLazyIMT.sol
+++ b/packages/imt.sol/contracts/internal/InternalLazyIMT.sol
@@ -157,6 +157,70 @@ library InternalLazyIMT {
         return _root(self, numberOfLeaves, depth);
     }
 
+    function _merkleProofElements(LazyIMTData storage self, uint40 index, uint8 depth)
+                                  internal view returns (uint256[] memory) {
+        uint40 numberOfLeaves = self.numberOfLeaves;
+        require(index < numberOfLeaves, "LazyIMT: leaf must exist");
+        require(depth > 0, "LazyIMT: depth must be > 0");
+        require(depth <= MAX_DEPTH, "LazyIMT: depth must be < MAX_DEPTH");
+
+        uint256[] memory proof = new uint256[](depth);
+        uint256 levelCount = self.numberOfLeaves;
+
+        for (uint8 i = 0; i < depth; ) {
+            // Left leaf
+            if (index & 1 == 0) {
+                bool outsideLimits = ((index+1) << i) % self.numberOfLeaves != 0;
+                if ((index + 1) < levelCount) {
+                    proof[i] = self.elements[_indexForElement(i, index + 1)];
+                } else if (((index+1) == levelCount) && (outsideLimits)) {
+                    proof[i] = _root(self, numberOfLeaves, i);
+                } else {
+                    proof[i] = _defaultZero(i);
+                }
+            // Right leaf
+            } else {
+                proof[i] = self.elements[_indexForElement(i, index -1)];
+            }
+            unchecked {
+                index >>= 1;
+                i++;
+            }
+            levelCount = numberOfLeaves >> i;
+        }
+        return proof;
+    }
+
+    function _merkleProofIndexes(LazyIMTData storage self, uint40 index, uint8 depth)
+                                 internal view returns (bool[] memory) {
+        uint40 numberOfLeaves = self.numberOfLeaves;
+        require(index < numberOfLeaves, "LazyIMT: leaf must exist");
+        require(depth > 0, "LazyIMT: depth must be > 0");
+        require(depth <= MAX_DEPTH, "LazyIMT: depth must be < MAX_DEPTH");
+
+        bool[] memory proofIndexes = new bool[](depth);
+
+        uint256 levelCount = self.numberOfLeaves;
+        for (uint8 i = 0; i < depth; ) {
+            if (levelCount > index >> 1) {
+                // Left leaf
+                if (index & 1 == 0) {
+                    proofIndexes[i] = false;
+                // Right leaf
+                } else {
+                    proofIndexes[i] = true;
+                }
+            }
+
+            unchecked {
+                index >>= 1;
+                i++;
+            }
+            levelCount = numberOfLeaves >> i;
+        }
+        return proofIndexes;
+    }
+
     function _root(LazyIMTData storage self, uint8 depth) internal view returns (uint256) {
         require(depth > 0, "LazyIMT: depth must be > 0");
         require(depth <= MAX_DEPTH, "LazyIMT: depth must be <= MAX_DEPTH");
--- a/packages/imt.sol/contracts/internal/InternalLazyIMT.sol
+++ b/packages/imt.sol/contracts/internal/InternalLazyIMT.sol
@@ -157,8 +157,11 @@ library InternalLazyIMT {
         return _root(self, numberOfLeaves, depth);
     }
 
-    function _merkleProofElements(LazyIMTData storage self, uint40 index, uint8 depth)
-                                  internal view returns (uint256[] memory) {
+    function _merkleProofElements(
+        LazyIMTData storage self,
+        uint40 index,
+        uint8 depth
+    ) internal view returns (uint256[] memory) {
         uint40 numberOfLeaves = self.numberOfLeaves;
         require(index < numberOfLeaves, "LazyIMT: leaf must exist");
         require(depth > 0, "LazyIMT: depth must be > 0");
@@ -170,17 +173,16 @@ library InternalLazyIMT {
         for (uint8 i = 0; i < depth; ) {
             // Left leaf
             if (index & 1 == 0) {
-                bool outsideLimits = ((index+1) << i) % self.numberOfLeaves != 0;
+                bool outsideLimits = ((index + 1) << i) % self.numberOfLeaves != 0;
                 if ((index + 1) < levelCount) {
                     proof[i] = self.elements[_indexForElement(i, index + 1)];
-                } else if (((index+1) == levelCount) && (outsideLimits)) {
+                } else if (((index + 1) == levelCount) && (outsideLimits)) {
                     proof[i] = _root(self, numberOfLeaves, i);
                 } else {
                     proof[i] = _defaultZero(i);
                 }
-            // Right leaf
             } else {
-                proof[i] = self.elements[_indexForElement(i, index -1)];
+                proof[i] = self.elements[_indexForElement(i, index - 1)];
             }
             unchecked {
                 index >>= 1;
@@ -191,8 +193,11 @@ library InternalLazyIMT {
         return proof;
     }
 
-    function _merkleProofIndexes(LazyIMTData storage self, uint40 index, uint8 depth)
-                                 internal view returns (bool[] memory) {
+    function _merkleProofIndexes(
+        LazyIMTData storage self,
+        uint40 index,
+        uint8 depth
+    ) internal view returns (bool[] memory) {
         uint40 numberOfLeaves = self.numberOfLeaves;
         require(index < numberOfLeaves, "LazyIMT: leaf must exist");
         require(depth > 0, "LazyIMT: depth must be > 0");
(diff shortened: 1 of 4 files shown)


alrevuelta:
@vimwitch @cedoor This PR is ready for review. I managed to use `_root` function to calculate the subhashes that either i) not calculated (odd leaf) or ii) require padding. Unit tests show this works for multiple depths and amount of leafs, and verifies that proofs are indeed valid with the already existing root functon.

packages/imt.sol/contracts/internal/InternalLazyIMT.sol
```diff
@@ -157,6 +157,74 @@ library InternalLazyIMT {
...
+        uint256 levelCount = self.numberOfLeaves;
+
+        for (uint8 i = 0; i < depth; ) {
+            // Left leaf
+            if (index & 1 == 0) {
+                bool outsideLimits = ((index + 1) << i) % self.numberOfLeaves != 0;
+                if ((index + 1) < levelCount) {
+                    proof[i] = self.elements[_indexForElement(i, index + 1)];
+                } else if (((index + 1) == levelCount) && (outsideLimits)) {
+                    proof[i] = _root(self, numberOfLeaves, i);
```
vimwitch, line 180: This function is calling into the `_root` function which is building the tree up to this level. This results in _many_ duplicate hashes being calculated, something like `O(N^2)` vs `O(N)`.

I think this function can be entirely removed and the elements of the `levels` array in `_root` can be returned directly. One approach to this could be splitting the logic from `_root` into a function that returns the levels, then in `_root` just returning the top level item, and in `_merkleProofElements` returning the levels.

Something like below should work and minimize duplicate code. The types/logic will need to be fixed but the structure should work.

```ts
function _root(LazyIMTData storage self, uint40 numberOfLeaves, uint8 depth) internal view returns (uint256) {
         // this should always short circuit if self.numberOfLeaves == 0
        if (numberOfLeaves == 0) return _defaultZero(depth);
        uint256 memory levels = _levels(self, numberOfLeaves, depth, numberOfLeaves - 1);
        return levels[depth];
}

function _merkleProofElements(LazyIMTData storage self, uint40 numberOfLeaves, uint8 depth, uint40 index) internal view returns (uint256) {
        uint256 memory levels = _levels(self, numberOfLeaves, depth, index);
        return levels;
}

function _levels(LazyIMTData storage self, uint40 numberOfLeaves, uint8 depth, uint40 index) internal view returns (uint256[] memory) {
        require(depth <= MAX_DEPTH, "LazyIMT: depth must be <= MAX_DEPTH");
       
        // accept as argument instead of hardcoding
        // uint40 index = numberOfLeaves - 1;
        uint256[MAX_DEPTH + 1] memory levels;
        if (index & 1 == 0) {
            levels[0] = self.elements[_indexForElement(0, index)];
        } else {
            levels[0] = _defaultZero(0);
        }
        for (uint8 i = 0; i < depth; ) {
            if (index & 1 == 0) {
                levels[i + 1] = PoseidonT3.hash([levels[i], _defaultZero(i)]);
            } else {
                uint256 levelCount = (numberOfLeaves) >> (i + 1);
                if (levelCount > index >> 1) {
                    uint256 parent = self.elements[_indexForElement(i + 1, index >> 1)];
                    levels[i + 1] = parent;
                } else {
                    uint256 sibling = self.elements[_indexForElement(i, index - 1)];
                    levels[i + 1] = PoseidonT3.hash([sibling, levels[i]]);
                }
            }
            unchecked {
                index >>= 1;
                i++;
            }
        }
        return levels;
}

```

packages/imt.sol/contracts/internal/InternalLazyIMT.sol
```diff
@@ -157,6 +157,74 @@ library InternalLazyIMT {
         return _root(self, numberOfLeaves, depth);
     }
 
+    function _merkleProofElements(
```
vimwitch, line 160: Please don't put this function between the `_root` overloads. This function can be either above or below the 3 `_root` functions.

packages/imt.sol/contracts/internal/InternalLazyIMT.sol
```diff
@@ -157,6 +157,74 @@ library InternalLazyIMT {
...
+            unchecked {
+                index >>= 1;
+                i++;
+            }
+            levelCount = numberOfLeaves >> i;
+        }
+        return proof;
+    }
+
+    function _merkleProofIndexes(
```
vimwitch, line 196: I don't think this function should exist in the contract. This logic can be calculated offchain like

```js
function indices(i, depth) {
  const out = []
  for (let x = 0; x < depth; x++) {
    out.push((i >> x) & 1)
  }
  return out
}
```

I think most root calculation functions should accept just the index of the leaf, and calculate intermediate indices internally.

packages/imt.sol/test/LazyIMT.ts
```diff
@@ -2,10 +2,26 @@ import { expect } from "chai"
 import { run } from "hardhat"
 import { poseidon2 } from "poseidon-lite"
 import { IMT } from "@zk-kit/imt"
+import type { BigNumber } from "ethers"
 import { LazyIMT, LazyIMTTest } from "../typechain-types"
 
 const random = () => poseidon2([Math.floor(Math.random() * 2 ** 40), 0])
 
+// Given a a merkle proof (elements and indexes) and a leaf, calculates the root
+function calculateRoot(element: BigInt, proofElements: BigNumber[], proofIndexes: boolean[]) {
```
vimwitch, line 11: This function should be inlined in the `Should produce valid Merke proofs for different trees` test because it is not used anywhere else.

packages/imt.sol/test/LazyIMT.ts
```diff
@@ -289,6 +305,47 @@ describe("LazyIMT", () => {
...
+                20: [9, 14, 15, 16, 18, 26, 27, 28, 40, 128, 129]
+            }
+
+            // For each depth
+            // eslint-disable-next-line guard-for-in
+            for (const depth in tests) {
+                // For each amount of leafs
+                for (const numLeaf of tests[depth]) {
+                    // Create the tree
+                    await lazyIMTTest.init(depth)
```
vimwitch, line 325: The `init` function cannot be called multiple times in a single test. The elements mapping will include entries from the previous tree. To safely do this you should take a [snapshot](https://hardhat.org/hardhat-network/docs/reference#evm_snapshot) before `init` then revert to the snapshot after executing the logic in the loop.

packages/imt.sol/contracts/internal/InternalLazyIMT.sol
alrevuelta: done 9e27c98ef556e3bf4c30e7ae716d126f274cd620

packages/imt.sol/contracts/internal/InternalLazyIMT.sol
alrevuelta: indeed, done: 9e27c98ef556e3bf4c30e7ae716d126f274cd620

packages/imt.sol/test/LazyIMT.ts
alrevuelta: right, moved inside: 9e27c98ef556e3bf4c30e7ae716d126f274cd620

packages/imt.sol/test/LazyIMT.ts
alrevuelta: done, using `evm_spapshoot` and `evm_revert`: 9e27c98ef556e3bf4c30e7ae716d126f274cd620

packages/imt.sol/contracts/internal/InternalLazyIMT.sol
alrevuelta: @vimwitch Acknowledge the problem and love the idea of reusing `_levels` for both `root` and `_merkleProofElements` but not sure this works as expected. Aren't the hashes different?

I mean, when going bottom->up, `root` calculates the hashes using the right or left element (either _defaultZero(x) or an existing leaf) but in `_merkleProofElements`, we are interested in the opposite leaf to construct the proof.

Am I missing something?

vimwitch: Yes i think you're right, the logic isn't exactly the same. I'd advise copying the logic from the `_root` function and changing it a bit. The merkle path function should be calling Poseidon directly. It might be possible to tweak such an internal `_levels` function to work for both, but not strictly necessary.

Also one other thing i forgot to mention, instead of calling it `MerkleTreeIndexes` i think `MerkleTreeIndices` is more correct.

packages/imt.sol/test/LazyIMT.ts
vimwitch: [Here](https://github.com/privacy-scaling-explorations/zk-kit/commit/9e27c98ef556e3bf4c30e7ae716d126f274cd620#diff-3545c4e73324d45a6e7ca34c851fb8d726dc945310c2412c7890808ef75ba6e4R324) you're snapshot/reverting outside of the `for` loop which makes it a noop. Each test gets it's own contract instance to work with, which is why we don't need snapshots in other tests. In this test we're initing inside a loop, so the snapshot/revert should happen inside that same loop.

packages/imt.sol/contracts/internal/InternalLazyIMT.sol
alrevuelta: > Also one other thing i forgot to mention, instead of calling it MerkleTreeIndexes i think MerkleTreeIndices is more correct.
right! done: 19c0b431ff44a779a46591d8d5516083214a8319

> Yes i think you're right, the logic isn't exactly the same. I'd advise copying the logic from the _root function and changing it a bit. The merkle path function should be calling Poseidon directly. It might be possible to tweak such an internal _levels function to work for both, but not strictly necessary.

@vimwitch Would this optimization be blocking for this PR? Not sure my other commitments would allow me to allocate time for this, or at least in a while.

packages/imt.sol/test/LazyIMT.ts
alrevuelta: far finger error, fixed: 19c0b431ff44a779a46591d8d5516083214a8319

vimwitch: I think you should create a new snapshot inside the loop as well. This seems to work but [the docs](https://hardhat.org/hardhat-network/docs/reference#evm_snapshot) say a single snapshot should not be reverted to more than once.

packages/imt.sol/contracts/internal/InternalLazyIMT.sol
alrevuelta: @vimwitch Note that for each tree I'm getting the merkle proof of all leafs, so that's around 700 calls to the `_merkleProofElements` in total.

Feel free to apply the changes, really curious about the results. Thanks!

vimwitch: Yes true, the test case is large.

I opened a pr alrevuelta/zk-kit#1. The approach i described mostly works. Each time merkle elements are loaded a merkle proof for the latest leaf is built to construct any subtrees that haven't been built yet. It should reduce the number of poseidon hashes that are required on average to get the elements, though the performance difference isn't notable in the current test cases.

Let me know what you think.

alrevuelta: Seems indeed way more efficient. Thanks!

packages/imt.sol/test/LazyIMT.ts
alrevuelta: done: fbf2c9c4e7d3ca114d314b26c52786abd99cec6e

packages/imt.sol/contracts/internal/InternalLazyIMT.sol
alrevuelta: Thanks for the help. This should solve this comment: https://github.com/alrevuelta/zk-kit/pull/1

alrevuelta:
Ready on my side. Added on top of the merged PR a fix for eslint.

alrevuelta:
@vimwitch Is there anything left here? Only [this](https://github.com/privacy-scaling-explorations/zk-kit/pull/162#discussion_r1502607619) seems unresolved.

vimwitch (approved):

Issue #131, opened 2024-01-22
Circuits - Extract useful logic from MACI


ctrlc03:
[**Is your feature request related to a problem? Please describe.**

[MACI](https://github.com/privacy-scaling-explorations/maci) contains a number of custom Circom circuits which could be re used by other zk developers. 

**Describe the solution you'd like**

This issue proposes to extract useful templates and integrate them with zk-kit, to allow more developers to easily access well documented and tested Circom code. 

List of templates:

- [x] ECDH - #130 
- [x] Poseidon - https://github.com/privacy-scaling-explorations/maci/pull/1169
- [x] SHA256 - https://github.com/privacy-scaling-explorations/maci/pull/1187
- [x] Unpack Element - #163 
- [x] Comparators - #164 
- [x] Float - #165 
- [ ] EdDSA Signatures verification (MACI rewrite)
- [ ] Merkle Trees (binary/quinary) (MACI rewrite)

> Ongoing [target branch](https://github.com/privacy-scaling-explorations/zk-kit/tree/feat/maci-circuits)

PR #163, opened 2024-02-13
Add `UnpackElement` circuit

0xjei:
## Description
This PR adds the `UnpackElement` circom template. This circuit allows a single field element to be converted into multiple 50-bit elements. 

Compared to the original version of MACI, Anonymous Components were used, resulting in DX improvement and small optimisation.

## Related Issue(s)
see #131 

## Checklist

### DIFF
--- a/packages/circuits/tests/unpack-element.test.ts
+++ b/packages/circuits/tests/unpack-element.test.ts
@@ -1,0 +1,47 @@
+import { WitnessTester } from "circomkit"
+import { circomkit } from "./common"
+
+describe("unpack-element", () => {
+    let circuit: WitnessTester<["in"], ["out"]>
+
+    it("Should unpack a field element with 3 / 4 / 5 packed values correctly", async () => {
+        for (let n = 3; n <= 5; n += 1) {
+            const elements: string[] = []
+
+            circuit = await circomkit.WitnessTester("unpack-element", {
+                file: "unpack-element",
+                template: "UnpackElement",
+                params: [n]
+            })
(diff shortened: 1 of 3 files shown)


ctrlc03 (approved):
Great stuff!!

packages/circuits/tests/unpack-element.test.ts
```diff
@@ -0,0 +1,47 @@
...
+            const INPUT = {
+                in: BigInt(`0b${elements.join("")}`)
+            }
+
+            const witness = await circuit.calculateWitness(INPUT)
+            await circuit.expectConstraintPass(witness)
+
+            const output = []
+            for (let i = 0; i < n; i += 1) {
+                const out = await getSignal(circuit, witness, `out[${i}]`)
```
cedoor, line 36: I'm not super sure about how this works. Just wanted to double-check. We are getting the output generated with `calculateWitness` right?

If so, doesn't the `expectPass` test always pass? Shouldn't it use an explicit value as the expected output rather than calculating it?

0xjei: See [8027bf7](https://github.com/privacy-scaling-explorations/zk-kit/pull/163/commits/8027bf7d81f4b5c207ef025d83e72195776f4cc9)

0xjei:
The test CI fails because the `poseidon-proof` zkeys have been updated and the verification keys do not match the new circuit model.

cedoor (approved):

PR #167, opened 2024-02-19
No implicit conversions in `packPoint`/`unpackPoint` functions

### LINKED ISSUE #154: Avoid implicit conversions to/from LE (Little Endian) in `packPoint`/`unpackPoint` functions
**Describe the bug**

The `packPoint` function in `@zk-kit/baby-jubjub` writes a `bigint` into a `buffer` in little-endian order, mutates it, then reads it back to a `bigint` using the default big-endian order. If that packed point is treated as an opaque object only usable by `unpackPoint`, this is fine since `unpackPoint` reverses the process. However, in order to be consistent `packPoint` should use `leBufferToBigint` instead of `bufferToBigint`, so that the packed points (as bigints) are the same between `@zk-kit/baby-jubjub` and `circomlibjs`.

**Additional context**

The discussion around this issue was started by @artwyman on Discord: https://discord.com/channels/943612659163602974/1181613897871007834/1204928272144539718 (#zk-kit channel of the PSE Discord).

cedoor:
## Description

This PR replaces the implicit conversions from LE to BE in the `packPoint`/`unpackPoint` functions with the utility function to convert in LE. More info and context in #154.

## Related Issue(s)

Fixes #154

## Checklist

### DIFF
--- a/packages/baby-jubjub/src/baby-jubjub.ts
+++ b/packages/baby-jubjub/src/baby-jubjub.ts
@@ -1,4 +1,4 @@
-import { F1Field, bigintToHexadecimal, bufferToBigint, leBigintToBuffer, leBufferToBigint, scalar } from "@zk-kit/utils"
+import { F1Field, leBigintToBuffer, leBufferToBigint, scalar } from "@zk-kit/utils"
 import * as sqrt from "./sqrt"
 import { Point } from "./types"
 
@@ -102,11 +102,11 @@ export function packPoint(unpackedPoint: Point<bigint>): bigint {
         buffer[31] |= 0x80
     }
 
-    return bufferToBigint(buffer)
+    return leBufferToBigint(buffer)
 }
 
 export function unpackPoint(packedPoint: bigint): Point<bigint> | null {
-    const buffer = Buffer.from(bigintToHexadecimal(packedPoint).padStart(64, "0"), "hex")
+    const buffer = leBigintToBuffer(packedPoint)
     const unpackedPoint = new Array(2)
 
     let sign = false
--- a/packages/baby-jubjub/tests/index.test.ts
+++ b/packages/baby-jubjub/tests/index.test.ts
@@ -1,3 +1,4 @@
+import { leBufferToBigint } from "@zk-kit/utils"
 import { babyjub } from "circomlibjs"
 import { Base8, Point, addPoint, inCurve, mulPointEscalar, packPoint, unpackPoint } from "../src"
 
@@ -34,7 +35,7 @@ describe("BabyJubjub", () => {
 
         const expectedPackedPoint = babyjub.packPoint(publicKey)
 
-        expect(packedPoint).toBe(BigInt(`0x${Buffer.from(expectedPackedPoint).toString("hex")}`))
+        expect(packedPoint).toBe(leBufferToBigint(expectedPackedPoint))
     })
 
     it("Should unpack a packed public key", async () => {
--- a/packages/baby-jubjub/tests/index.test.ts
+++ b/packages/baby-jubjub/tests/index.test.ts
@@ -32,9 +32,11 @@ describe("BabyJubjub", () => {
 
     it("Should pack a point", async () => {
         const packedPoint = packPoint(publicKey)
+        const strippedPackedPoint = packedPoint & ~(BigInt(1) << BigInt(255))
 
         const expectedPackedPoint = babyjub.packPoint(publicKey)
 
+        expect(strippedPackedPoint).toBe(publicKey[1])
         expect(packedPoint).toBe(utils.leBuff2int(expectedPackedPoint))
     })
 
--- a/packages/baby-jubjub/tests/index.test.ts
+++ b/packages/baby-jubjub/tests/index.test.ts
@@ -32,6 +32,9 @@ describe("BabyJubjub", () => {
 
     it("Should pack a point", async () => {
         const packedPoint = packPoint(publicKey)
+        // As a bigint, we expect the packed point to be identical to the Y coordinate,
+        // except for the 1 bit added to represent whether the X coordinate is negative or positive.
+        // We strip off that extra bit and check this expectation below.
         const strippedPackedPoint = packedPoint & ~(BigInt(1) << BigInt(255))
 
         const expectedPackedPoint = babyjub.packPoint(publicKey)
--- a/packages/baby-jubjub/tests/index.test.ts
+++ b/packages/baby-jubjub/tests/index.test.ts
@@ -1,5 +1,5 @@
-import { leBufferToBigint } from "@zk-kit/utils"
 import { babyjub } from "circomlibjs"
+import { utils } from "ffjavascript"
 import { Base8, Point, addPoint, inCurve, mulPointEscalar, packPoint, unpackPoint } from "../src"
 
 describe("BabyJubjub", () => {
@@ -35,7 +35,7 @@ describe("BabyJubjub", () => {
 
         const expectedPackedPoint = babyjub.packPoint(publicKey)
 
-        expect(packedPoint).toBe(leBufferToBigint(expectedPackedPoint))
+        expect(packedPoint).toBe(utils.leBuff2int(expectedPackedPoint))
     })
 
     it("Should unpack a packed public key", async () => {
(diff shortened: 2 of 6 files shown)


packages/baby-jubjub/tests/index.test.ts
```diff
@@ -34,7 +35,7 @@ describe("BabyJubjub", () => {
 
         const expectedPackedPoint = babyjub.packPoint(publicKey)
```
artwyman, line 36: Might make sense to convert this to a bigint (using Scalar.fromRprLE) and compare as bigints.  Not really different in functionality, but might better match the intent that the output should be "the same number" with each library converting in its own preferred way.

The direct use of Scalar.fromRprLE makes this more awkward than ideal, so this is a weak suggestion rather than a must-do.  It would be more ideal if this could be a call to toObject(), which is more obviously "circomlibjs's way of converting buffers to numbers for public consumption".  However toObject() explicitly doesn't work for packed points because it tries to convert from Montgomery form, and that's already been done during packing.

packages/baby-jubjub/src/baby-jubjub.ts
```diff
@@ -102,11 +102,11 @@ export function packPoint(unpackedPoint: Point<bigint>): bigint {
         buffer[31] |= 0x80
     }
 
-    return bufferToBigint(buffer)
+    return leBufferToBigint(buffer)
 }
 
 export function unpackPoint(packedPoint: bigint): Point<bigint> | null {
```
artwyman, line 108: You can decide on the right best-practices for your library, but I'd suggest that functions like this have more explicit documentation which can explain what a packed point is, and what assumptions users should or shouldn't make about it.  I don't know if this exists, but is there a standard definition of what point packing means, or a compatibility target which could be pointed to here?

artwyman (commented):
I don't know if there are uses of these functions outside of zk-kit which will need to be updated for this change.  Do you have a process for checking compatibility, and for reflecting breaking changes in release notes, version numbers, etc?

artwyman:
These changes reflect the functional changes we discussed.  My review comments are mostly suggestions around clarity, and questions about the release process on this repo, which I don't understand yet.  I'm willing to change my review to an Accept if you want to move forward quickly.

packages/baby-jubjub/tests/index.test.ts
cedoor: Correct me if I'm wrong but I remember there was no conversion to/from Montgomery in `circomlibjs@0.0.8`.

I think it's still worth using `Scalar.fromRprLE` to really test if the two libraries match their outputs with *their* functions.

cedoor: I actually used: `leInt2Buff`, to match what `circomlibjs` uses here: https://github.com/iden3/circomlibjs/blob/c71c8c1daf54d76db8e0858606faa017783df6a1/src/babyjub.js#L97.

Let me know if it makes sense!

packages/baby-jubjub/src/baby-jubjub.ts
cedoor: I don't know if there's a standard but we can definitely look for it!

Documenting these libraries is part of a separate task: https://github.com/privacy-scaling-explorations/zk-kit/issues/142

We definitely need to add more documentation and tests too. The current version is more like an alpha. The goal of this [milestone](https://github.com/orgs/privacy-scaling-explorations/projects/2/views/4) is to release `1.0.0-beta`, to be ready for being audited by the internal PSE team.

cedoor:
> I don't know if there are uses for these functions outside of zk-kit which will need to be updated for this change. Do you have a process for checking compatibility, and for reflecting breaking changes in release notes, version numbers, etc?

For the time being to my knowledge it is only used by Semaphore and MACI. We have not yet integrated an automatic release mechanism into zk-kit, but the idea is to use tools such as [changelogithub](https://github.com/antfu/changelogithub), which is based on conventional commits (`yarn commit` forces you to meet that convention). As you can see we're starting to organize this repo now. Will have clearer processes soon.

packages/baby-jubjub/tests/index.test.ts
artwyman: Great point.  All my digging into circomlibjs code has been based on the latest version, so needs to be adjusted for the different version you're using.

artwyman:
> For the time being to my knowledge it is only used by Semaphore and MACI. We have not yet integrated an automatic release mechanism into zk-kit, but the idea is to use tools such as [changelogithub](https://github.com/antfu/changelogithub), which is based on conventional commits (`yarn commit` forces you to meet that convention). As you can see we're starting to organize this repo now. Will have clearer processes soon.

Sounds reasonable.  As I'm starting to use zk-kit in my prototyping, I'm leaving all the version numbers pinned with no assumption of accepting future versions (e.g. `"@zk-kit/baby-jubjub": "0.1.1"` not `^0.1.1`), so I'll have a chance to adjust code when new versions change behavior like this.

packages/baby-jubjub/tests/index.test.ts
```diff
@@ -34,7 +35,7 @@ describe("BabyJubjub", () => {
...
 
-        expect(packedPoint).toBe(BigInt(`0x${Buffer.from(expectedPackedPoint).toString("hex")}`))
+        expect(packedPoint).toBe(utils.leBuff2int(expectedPackedPoint))
```
artwyman, line 38: Optionally, I thought of another good test check which might be worth adding, which would directly reflect the mismatch which caused me to detect this in the first place.

As a bigint, we expect the packed point to be identical to the Y coordinate, except for the 1 bit added to represent the whether the X coordinate is negative or positive.  You could strip off that extra bit and check this expectation like this (syntax may not be perfect, I haven't tried compiling this):

const strippedPackedPoint = packedPoint & ~(1n<<255n);
expect(strippedPackedPoint).toBe(publicKey[0]);

You could also check that the high bit is as-expected based on the value of publicKey[1], but that's unrelated to the byte order issue.

artwyman (approved):
One more suggestion in the test, but I think this looks good to me functionally.

packages/baby-jubjub/tests/index.test.ts
cedoor: Yess, makes sense! I've added it 👍🏽 The coordinate to check is `publicKey[0]` though.

artwyman: Right, I got array indexes wrong in my comment above.  My reference to Y and X was correct.

Side note based on something I learned about today: I often get this confused, and I think it's because the standard form of elliptic curves I've seen in examples is symmetric when reflected across the X axis, not the Y, which doesn't match what's being done here.  I think the important distinction is that the Baby-Jubjub library is using Edwards form of its curve, which seems to be symmetric across both X and Y (https://en.wikipedia.org/wiki/Edwards_curve).  Though it's symmetric on both axes, that article suggests that the definition of the opposite of a point on an Edwards curve by negating X.  That's probably why packPoint is defined the way it is here.

packages/baby-jubjub/tests/index.test.ts
```diff
@@ -31,10 +32,12 @@ describe("BabyJubjub", () => {
 
     it("Should pack a point", async () => {
         const packedPoint = packPoint(publicKey)
+        const strippedPackedPoint = packedPoint & ~(BigInt(1) << BigInt(255))
```
artwyman, line 35: I feel like what's being checked here is tricky enough to be worth a comment explaining what it's doing.

cedoor: Right, let me add it!

packages/baby-jubjub/tests/index.test.ts
cedoor: >  I think the important distinction is that the Baby-Jubjub library is using Edwards form of its curve, which seems to be symmetric across both X and Y (https://en.wikipedia.org/wiki/Edwards_curve). Though it's symmetric on both axes, that article suggests that the definition of the opposite of a point on an Edwards curve by negating X. That's probably why packPoint is defined the way it is here.

It should be like that. I've discussed this aspect with many people more expert than me and I got a rough idea. This should confirm that idea 🤔

ctrlc03 (approved):
Nothing more to add, great review by @artwyman and good work!

Issue #144, opened 2024-02-08
Make `babyjubjub`/`utils` packages browser-compatible


cedoor:
**Describe the improvement you're thinking about**

`Buffer` is an object only available on NodeJS. Although many bundlers add polyfills by default, some require configuration. Rollup itself could add a polyfill to the bundles, a browser-compatible package (e.g. [`buffer`](https://www.npmjs.com/package/buffer)) could be used, or `Buffer` could be replaced by `Uint8Array`, which is supported by both browsers and nodeJS.

artwyman:
Context from the Zupass repo (which may be helpful, though you don't have to follow our lead): We polyfill Buffer, to the point that I didn't even realize it wasn't in browsers until recently.  There's also a "buffer" NPM package we've imported to provide a compatible version, which may be an alternative to polyfill or eliminating use of Buffer.

I've found Buffer nice to work with compared to UInt8Array, but it's pretty easy to convert between them whenever I need to if zk-kit starts to return only UInt8Array.

cedoor:
Thanks! I think all these options need to be considered. The easiest solution looks like adding a polyfill. We're already using a Rollup plugin for the `iife` bundle here: https://github.com/privacy-scaling-explorations/zk-kit/blob/0b4ccbea88168d01cc518ed9b537f3479d6f7759/packages/utils/rollup.iife.config.ts#L38. We would just need to use in the other rollup configuration as well.

Replacing the `Buffer` functions we're using with `Uint8Array` could make the bundle lighter, but it also needs much more work and could introduce bugs.

On second thought, considering that in the future we may need other functions that `Buffer` already has, perhaps it is better to use a polyfill.

PR #175, opened 2024-02-21
Add `be` prefix & make some functions more generic

### LINKED ISSUE #153: Add `be` (Big Endian) prefix to the relevant conversion functions in the `utils` package

### LINKED ISSUE #174: Buffers returned by `leBigintToBuffer` and `beBigintToBuffer` should have dynamic length

The `leBigintToBuffer` and `beBigintToBuffer` functions only return buffers with 32 bytes. `leBigintToBuffer` was originally used inside the `eddsa-poseidon` package, where it only needed to return a buffer with a static size (32 bytes). Since those functions are part of the `utils` now, they should be generic and support different options. This is a breaking change as it will be mandatory to define the bytes if one needs to convert a number to a buffer with static length.

These functions should add an additional optional parameter: `length`. If `length` is not defined or if it is less than the number of bytes of the `bigint` to be converted, the minimum number of bytes required for the conversion will be used.  

Example:
```ts
export function beBigintToBuffer(n: bigint, length?: number): Buffer {
    const hex = bigIntToHexadecimal(n)

    // Calculate the minimum buffer length required to represent 'n' in bytes.
    // Each hexadecimal character represents 4 bits, so 2 characters are 1 byte.
    const minLength = Math.ceil(hex.length / 2)

    // Use the provided length or the calculated minimum length, whichever is greater.
    const bufferLength = length ? Math.max(length, minLength) : minLength

    // Allocate buffer of the desired size, filled with zeros.
    const buffer = Buffer.alloc(bufferLength, 0)

    const fromHex = Buffer.from(hex, "hex")
    fromHex.copy(buffer, bufferLength - fromHex.length)

    return buffer
}
```

cedoor:
## Description

This PR mainly covers the functions of the `conversions` module of the `@zk-kit/utils` library. It essentially does the following:

1. Adds the prefix `be` to functions that do conversions with hexadecimals or buffers in big-endian.
2. Renames the `Bigint`/`bigint` occurrences in function names to `BigInt`/`bigInt`.
3. Adds a `size` parameter to the `leBigintToBuffer` and `beBigintToBuffer` functions to define the number of bytes of the buffer returned by those function.
4. Adds tests for the module's `conversions` functions.
5. Documents all the code in the `conversions` module.

This PR introduces breaking changes! See the related issues for more info.

## Related Issue(s)

Closes #153
Closes #174

## Checklist


packages/utils/src/conversions.ts
```diff
@@ -0,0 +1,202 @@
...
+ * It also includes variations for both big-endian (`be`) and little-endian
+ * (`le`) conversions. It is important to note that when there is no prefix,
+ * the order of bytes is always big-endian.
+ */
+
+import { BigNumberish } from "./types"
+import { isHexadecimal, isStringifiedBigint } from "./number-checks"
+
+/**
+ * Converts a bigint to a hexadecimal string ensuring even length.
```
artwyman, line 19: If the function name makes byte-order explicit, the documentation should say what the byte order is, in order to describe what "be" stands for.  E.g. "Converts a bigint to a hexadecimal string ensuring even length.  Uses big-endian byte order."

This same comment applies to all of the functions in this file, so I'm not going to repeat it.  They should all make byte order explicit in the documentaiton.

packages/utils/src/conversions.ts
```diff
@@ -0,0 +1,202 @@
...
+
+    return hex
+}
+
+/**
+ * Converts a bigint to a hexadecimal string. Alias for beBigIntToHexadecimal.
+ * @param n The bigint value to convert.
+ * @returns The hexadecimal representation of the bigint.
+ */
+export function bigIntToHexadecimal(n: bigint): string {
```
artwyman, line 39: For hexadecimal stringification of a number, I actually don't think byte order needs to be explicit.  It doesn't hurt, but I think in this case there actually is an obvious default.  Numbers represented as strings always put the most significant digit first.  Importantly these functions are really outputting digits (4 bits at a time) not bytes.

Relatedly, in this case it's weird to make BE byte order explicit, but not provide any LE alternatives.

packages/utils/src/conversions.ts
```diff
@@ -0,0 +1,202 @@
...
+/**
+ * Converts a bigint to a hexadecimal string ensuring even length.
+ * @param n The bigint value to convert.
+ * @returns The hexadecimal representation of the bigint.
+ */
+export function beBigIntToHexadecimal(n: bigint): string {
+    let hex = n.toString(16)
+
+    // Ensure even length.
+    if (hex.length % 2 !== 0) {
```
artwyman, line 27: I don't know all the use cases, but I wonder if what they really need isn't padding to an even number, but padding out to a specific size (e.g. 32 bytes) which should be passed as an argument.  This doesn't need to be resolved in this PR, it's just something which occurs to me as an easy misuse.

packages/utils/src/conversions.ts
```diff
@@ -0,0 +1,202 @@
...
+ * Converts a bigint to a hexadecimal string. Alias for beBigIntToHexadecimal.
+ * @param n The bigint value to convert.
+ * @returns The hexadecimal representation of the bigint.
+ */
+export function bigIntToHexadecimal(n: bigint): string {
+    return beBigIntToHexadecimal(n)
+}
+
+/**
+ * Converts a hexadecimal string to a bigint, ensuring the hex string starts with '0x'.
```
artwyman, line 44: "ensuring the hex string starts with 0x" isn't behavior which is visible to the caller, so it doesn't seem like the right thing to call out in documentation.

Alternative suggestions: "Coverts a hexadecimal string to a bigint.  The input is interpreted as hexadecimal with or without a `0x` prefix."

packages/utils/src/conversions.ts
```diff
@@ -0,0 +1,202 @@
...
+ * Converts a hexadecimal string to a bigint. Alias for beHexadecimalToBigint.
+ * @param hex The hexadecimal string to convert.
+ * @returns The bigint representation of the hexadecimal string.
+ */
+export function hexadecimalToBigint(hex: string): bigint {
+    return beHexadecimalToBigint(hex)
+}
+
+/**
+ * Converts a buffer to a bigint by interpreting the buffer as a hexadecimal string.
```
artwyman, line 65: "by interpreting the buffer as a hexadecimal string" is an implementation detail which I think is actually confusing here.  It suggests to me that as caller I should be passing an array of strings.  The buffer isn't being intereted as a hex string, it's being converted to a hex string internally via the call to toString() below.

Alternative suggestion: "Converts a buffer of bytes to a bigint using big-endian byte order."

packages/utils/src/conversions.ts
```diff
@@ -0,0 +1,202 @@
...
+ * Converts a buffer to a bigint. Alias for beBufferToBigint.
+ * @param b - The buffer to convert.
+ * @returns The bigint representation of the buffer's contents.
+ */
+export function bufferToBigint(b: Buffer): bigint {
+    return beBufferToBigint(b)
+}
+
+/**
+ * Converts a bigint to a buffer, assuming big-endian byte order, and fills with zeros if necessary.
```
artwyman, line 83: I'd say "using" rather than "assuming" big-endian byte order.

"assuming" to me would suggest that big-endian is a property of the bigint, while it's actually a property of the buffer introduced by the conversion.

Same comment applies to the other uses of "assuming" in this file, so I won't repeat it.

packages/utils/src/conversions.ts
```diff
@@ -0,0 +1,202 @@
...
+ * @returns The buffer representation of the bigint.
+ */
+export function beBigintToBuffer(n: bigint, length?: number): Buffer {
+    const hex = bigIntToHexadecimal(n)
+
+    // Calculate the minimum buffer length required to represent 'n' in bytes.
+    // Each hexadecimal character represents 4 bits, so 2 characters are 1 byte.
+    const minLength = Math.ceil(hex.length / 2)
+
+    // Use the provided length or the calculated minimum length, whichever is greater.
```
artwyman, line 95: This behavior needs to be documented or it'll be surprising.  Perhaps the argument should be called minLenth?
Alternatively if minLength > length this function could throw an exception.  Incorrect buffer lengths can be a source of security bugs, so I think checking and failing loudly might be the best approach.  (The current approach is already much safer than truncating, though.)

packages/utils/src/conversions.ts
```diff
@@ -0,0 +1,202 @@
...
+    return beBigintToBuffer(n)
+}
+
+/**
+ * Converts a BigNumberish type to a buffer, handling various input types and converting them to bigint first if necessary.
+ * @param n The BigNumberish value to convert.
+ * @returns The buffer representation of the BigNumberish value.
+ */
+export function beBigNumberishToBuffer(n: BigNumberish): Buffer {
+    if (
```
artwyman, line 122: The logic here could be delegated to `bigNumberishToBigint` to avoid duplication. That would simplify this function to only need one special case for input which is already a buffer.

  if (typeof n === "Buffer") {
    return n as Buffer;
  }
  return bigintToBuffer(bignumberishToBigint(n));

packages/baby-jubjub/src/baby-jubjub.ts
```diff
@@ -1,4 +1,4 @@
-import { F1Field, bigintToHexadecimal, bufferToBigint, leBigintToBuffer, leBufferToBigint, scalar } from "@zk-kit/utils"
+import { F1Field, bigIntToHexadecimal, bufferToBigint, leBigintToBuffer, leBufferToBigint, scalar } from "@zk-kit/utils"
```
artwyman, line 1: This name change seems inconsistent with other functions in this library.  E.g. leBigintToBuffer doesn't capitalize the `I` in Bigint, so why should this one function do so?

It would be good to standardize on one rule for whether that I should be capitalized.

packages/utils/src/conversions.ts
```diff
@@ -0,0 +1,202 @@
...
+export function bigintToBuffer(n: bigint): Buffer {
+    return beBigintToBuffer(n)
+}
+
+/**
+ * Converts a BigNumberish type to a buffer, handling various input types and converting them to bigint first if necessary.
+ * @param n The BigNumberish value to convert.
+ * @returns The buffer representation of the BigNumberish value.
+ */
+export function beBigNumberishToBuffer(n: BigNumberish): Buffer {
```
artwyman, line 121: These "bignumberish" helpers are another case where it's odd not to have LE alternatives.  Unlike the hex digits above, I do think this is a case where it's good to make the byte order explicit, since the use of a buffer means there is a choice.

packages/utils/src/conversions.ts
```diff
@@ -0,0 +1,202 @@
...
+ * @returns The buffer representation of the bigint in little-endian.
+ */
+export function leBigintToBuffer(n: bigint, length?: number): Buffer {
+    const hex = bigIntToHexadecimal(n)
+
+    // Calculate the minimum buffer length required to represent 'n' in bytes.
+    // Each hexadecimal character represents 4 bits, so 2 characters are 1 byte.
+    const minLength = Math.ceil(hex.length / 2)
+
+    // Use the provided length or the calculated minimum length, whichever is greater.
```
artwyman, line 192: Similar to above, this behavior should be documented, and I have a similar question as to whether it's better to rename length -> minLength, or to throw on overflow.

packages/utils/tests/conversions.test.ts
```diff
@@ -0,0 +1,98 @@
...
+        it("Should convert a bigint to a BE hexadecimal", async () => {
+            const result = hexadecimalToBigint(testHex1BE)
+
+            expect(result).toBe(testBigInt1BE)
+        })
+    })
+
+    describe("# bigNumberishToBuffer", () => {
+        it("Should convert a big numberish to a BE buffer", async () => {
+            const result = bigNumberishToBuffer(testBigInt1BE)
```
artwyman, line 40: Should have tests exercising the other possible input types.

packages/utils/tests/conversions.test.ts
```diff
@@ -0,0 +1,98 @@
...
+
+            expect(result).toStrictEqual(Buffer.from(testBytes1.slice(1)))
+        })
+    })
+
+    describe("# bigNumberishToBigint", () => {
+        it("Should convert a BE big numberish to a bigint", async () => {
+            const result = bigNumberishToBigint(Buffer.from(testBytes1))
```
artwyman, line 48: Again, testing of the other input types would be welcome.

packages/utils/tests/conversions.test.ts
```diff
@@ -0,0 +1,98 @@
...
+
+            expect(result).toBe(testBigInt1BE)
+        })
+    })
+
+    describe("Bigint to/from Buffer Conversions", () => {
```
artwyman, line 54: Consider adding tests to exercise the overflow behavior I commented on elsewhere (i.e. what happens when the bigint is too big for the specified buffer size).  Your code-coverage analysis will probably point that out as an untested branch.

artwyman (changes requested):
Looks good overall.  A lot of comments primary around documentation and testing.

packages/utils/src/conversions.ts
cedoor: Right! I've added an explanation in the module description but it would be better to be explicit in those functions too 👍🏽

packages/utils/src/conversions.ts
cedoor: From what I know `Buffer` needs it. `Buffer.from("02", "hex")` is not the same as `Buffer.from("2", "hex")`.

It shouldn't make any difference in the other use-cases.

Feel free to open another issue for that feature if you want 👍🏽

cedoor: BTW there's an issue for replacing `Buffer` with `Uint8Array`, to make these libraries fully-compatible with browsers -> https://github.com/privacy-scaling-explorations/zk-kit/issues/144.

Your input would be really appreciated. I'll work on that asap.

packages/utils/src/conversions.ts
cedoor: I see. I'd tend to prefer leaving it without any prefix then. If we realize someone needs this conversion with little-endian byte order we could make it more explicit. What do you think?

packages/utils/src/conversions.ts
cedoor: Right, definitely clearer! Thanks

packages/utils/src/conversions.ts
cedoor: Good catch! The following code should be fine:
```ts
 if (n instanceof Buffer) {
        return n
 }

return bigintToBuffer(bigNumberishToBigint(n))
```

I'd like to add more checks on types here. But it'll be part of another PR: https://github.com/privacy-scaling-explorations/zk-kit/pull/178 (feel free to start taking a look).

packages/baby-jubjub/src/baby-jubjub.ts
cedoor: Right! IMO we should use "BigInt" everywhere, as it's the same name JS uses.

cedoor: I've renamed them. I started marking breaking changes in my commits: https://github.com/privacy-scaling-explorations/zk-kit/pull/175/commits/6858f66f6778bb9656ca97a391f8128ff354d0bf.

packages/utils/src/conversions.ts
cedoor: Makes sense! I've added `leBigNumberishToBuffer`, but not sure about `beBigNumberishToBigInt`.

packages/utils/src/conversions.ts
artwyman: I agree.

packages/utils/src/conversions.ts
artwyman: Why not?  I feel like there should be symmetry.

packages/utils/src/conversions.ts
```diff
@@ -5,7 +5,7 @@
  * conversions involving bigints, hexadecimals and buffers.
  * The module is structured with clear function naming to indicate
  * the conversion direction (e.g., `bigIntToHexadecimal` for BigInt
- * to hexadecimal, `bufferToBigint` for buffer to bigint) and employs
+ * to hexadecimal, `bufferToBigInt` for buffer to bigint) and employs
```
artwyman, line 8: I somehow missed this module documentation on my first review.  I think I'm used to skipping over big comment blocks at the tops of files which are usually copyright text.

Regardless, I like this description - very clear.

packages/utils/src/conversions.ts
```diff
@@ -41,111 +42,128 @@ export function bigIntToHexadecimal(n: bigint): string {
...
+    return beBufferToBigInt(b)
 }
 
 /**
- * Converts a bigint to a buffer, assuming big-endian byte order, and fills with zeros if necessary.
+ * Converts a bigint to a buffer and fills with zeros if necessary.
+ * It uses big-endian byte order.
  * @param n The bigint to convert.
- * @param length The number of bytes of the buffer to return.
+ * @param size The number of bytes of the buffer to return.
```
artwyman, line 97: You should document what happens if this parameter is absent.  Also document the error thrown if the size is too small.
I like the new version of the argument and logic, thanks.

packages/utils/src/conversions.ts
```diff
@@ -41,111 +42,128 @@ export function bigIntToHexadecimal(n: bigint): string {
...
 }
 
 /**
- * Converts a bigint to a buffer. Alias for beBigintToBuffer.
- * @param n - The bigint to convert.
- * @returns The buffer representation of the bigint.
+ * Converts a bigint to a buffer and fills with zeros if necessary.
+ * It uses little-endian byte order.
+ * @param n The bigint to convert.
+ * @param size The number of bytes of the buffer to return.
```
artwyman, line 126: Same need for documentation here of absent size, and error behavior.

artwyman (approved):
Looks good, thanks.  A few more suggestions in comments, but they're minor, so not going to block you on another round of review from me across timezones.

ctrlc03 (approved):

packages/utils/src/conversions.ts
cedoor: I also skit it usually. It will be more visible on the TypeDoc website I guess.

packages/utils/src/conversions.ts
cedoor: Right, a bignumber-ish could also be a buffer indeed.

artwyman: The case where the BigNumberish is itself a buffer is a tricky one here.  It raises the question of whether the BE/LE prefix applies to the input buffer, or the output buffer, or both.  I think it's too complicated to introduce 4 separate options to separately specify input and output byte order.  So I could see two approaches which could be consistent here:

1) The LE/BE specifier applies to **both** input and output.  That means when a conversion is needed, use the selected byte order.  But if the input is already a buffer, it's assumed to already be in the desired byte order, so it can be passed through unchanged.
2) The LE/BE specifier applies to the "Buffer" output.  BigNumberish is assumed (and should be documented) to always assume a BE buffer.  In this case, if the input is a Buffer you'd need to reverse it as you say.

I think #1 is the simplest option for this function.  It's really an argument justifying how it already works.  I think #2 might be more consistent broadly, because all the functions which accept a BigNumberish are doing BE conversion by default.

Realizing this complexity makes me feel more like it's best to not provide a LE version after all.  I'd just leave the function as-is, and document what it does for clarity.  In particular, document that if the input is a buffer it's left unmodified, and if the input is not a buffer it's converted using BE order.  So long as callers know that, they can figure out what they need to do.

cedoor: > Realizing this complexity makes me feel more like it's best to not provide a LE version after all. I'd just leave the function as-is, and document what it does for clarity.

I agree. I removed the LE version and added more documentation to explain how they work.

PR #178, opened 2024-02-21
New utility modules for handling errors and checking types

### LINKED ISSUE #177: New utility functions for checking types

Semaphore recently added a new package with some utility functions. The first modules in the package, which allow type checking and type error handling, can however also be used outside Semaphore. ZK-Kit might be a better place for that kind of function. 

Those modules could just be moved to the `@zk-kit/utils` package with the same structure. `@zk-kit/utils` could also provide conditional exports to allow devs to use some modules selectively as Semaphore did. 

**Additional context**

Semaphore PR: https://github.com/semaphore-protocol/semaphore/pull/643.

cedoor:
## Description

This PR adds two utility modules to the `@zk-kit/package`: `error-handlers` and `type-checks`. It also renames some modules and the `isStringifiedBigint` function to `isStringifiedBigInt`, which can be considered as a breaking change.

## Related Issue(s)

Closes #177

## Checklist


vplasencia (approved):

packages/eddsa-poseidon/src/eddsa-poseidon.ts
```diff
@@ -74,14 +69,14 @@ export function derivePublicKey(privateKey: BigNumberish): Point<string> {
  */
 export function signMessage(privateKey: BigNumberish, message: BigNumberish): Signature<string> {
     // Convert the private key to buffer.
-    privateKey = utils.checkPrivateKey(privateKey)
+    privateKey = checkPrivateKey(privateKey)
```
artwyman, line 72: Not strictly related to this PR, and I'm not sure if it's actionable, but I wanted to share some feedback about my experience with this.  I think the level of flexibility on inputs here has a downside that the caller may not always be sure how the input is being interpreted, which could have an impact for compatibility.

For example, in my tests using this library, I was using a test private key borrowed from here: https://github.com/iden3/circomlibjs/blob/4f094c5be05c1f0210924a3ab204d8fd8da69f49/test/eddsa.js#L103

That key is intended to be a hex string representing 32 bytes of key to be hashed by the initial blake hash below.  That's how I need to interpret it to remain consistent with existing uses of EdDSA in Zupass.  However the hex digits happen to all be in the range 0-9, so it was being interpreted as a stringified bigint.  This worked but was treating my string as a different key than I intended.  If any of my digits had been A-F, it would've instead treated my key as a raw string of 64 bytes, rather than 32 bytes, which again would've silently worked, but been treating it as a different key.  What I should have done (and will now do) is convert my string into a buffer in advance using `Buffer.from(pk, "hex")`, meaning I'll be bypassing all the type flexibility here.

As I said, I'm not sure if there's a proposal here, but I suggest thinking about the tradeoff between accepting flexible input, vs. being stricter in order to make sure users are only getting the behavior they expect.  Given that this function also accepts arbitrary strings (like passwords?), I think it might be best not to accept stringified bigints too, since it's impossible to know what the user's intent if the string happens to be made up of digits.  This is the kind of behavior which might be difficult to maintain compatibly if the code ever changes.

artwyman: There's actually another related point which I think is a potential bug.  If I put a "0x" before my private key, it's interpreted as hex, but ends up as a 31-byte buffer rather than a 32-byte buffer.  The reason is that the hex string starts with `00`, and the conversion from string -> bigint -> buffer ends up truncating the leading zeroes.  I'm not sure what you think is "correct" here, but if you intend bigints to be converted into fixed-size buffers, you should pass a size parameter here.

artwyman (changes requested):
I reviewed the code, but not configuration or tests.  I have a few small suggestions to resolve in this PR.  My comments also raise some more general concerns about the way flexible inputs may lead to bugs, but that discussion can be taken outside of this PR if you prefer.

packages/eddsa-poseidon/src/eddsa-poseidon.ts
artwyman: Actually I discovered that the conversion issue I mentioned was only in my test code, not my intended production library, which was already calling `Buffer.from(pk, "hex")`.  My point about the potential for confusion and mistakes remains, though it seems I don't have a major issue to fix on my side.  I'm going to be writing more compatibility tests to rule out this kind of issue in future.

cedoor: > As I said, I'm not sure if there's a proposal here, but I suggest thinking about the tradeoff between accepting flexible input, vs. being stricter in order to make sure users are only getting the behavior they expect.

I totally agree and I'd propose the following types: arbitrary strings (yes like passwords), `bigint`, `number`, `Buffer`, or hex strings (must start with '0x'). This is something we should think about now, before making this lib stable IMO. Let me know what you think.

cedoor: >  The reason is that the hex string starts with 00, and the conversion from string -> bigint -> buffer ends up truncating the leading zeroes.

The problem here is the way `bigNumberishToBuffer` converts hexadecimal strings to buffers. It could convert it by using `Buffer.from(n, "hex")` if it's a hex string. "0x" should be removed before converting it ofc. IMO if the hex starts with zeroes it should consider them when converting to buffer. How does it sound?

artwyman: I've thought through this a little more, so let me share my thought process.  I think the confusion is specific to this private key use case, not all the other Bignumberish use cases because what you want as input here isn't a single number.  As I understand it,  all (or at least most) of the Bignumberish inputs are actually representing a single number which is expected to fit in one field element (32 bytes, actually limited to 254 bits).  In that context, it's okay to accept a string as a convenience, and to convert it into a number in a reasonable way.  The use of the 0x prefix serves as an unambiguous way to decide whether to convert the string using decimal or hex.

In this case, thanks to the blake hash, the private key can really be any arbitrary set of bytes.  I don't know how that usage fits with what the EdDSA standard says, but any arbitrary bytes can work here.  A string can be treated as arbitrary bytes, though it could also be treated as hex bytes, which is still subtly different (due to byte order and truncation) from converting the string into a number.  If the string is a password, you can't make any assumptions about it's content, and you probably don't want it to accidentally behave differently if the pasword happens to be "1234" or "0xdeadbeef".  If you're treating passwords as bytes, the Bignumberish concept actually confuses things, since it introduces multiple interpretations of the same type, and is based on an assumption (we want one field element) which doesn't hold here.

With that in mind, I have two separate suggestions:

1) I think the most unambiguous thing to do for private keys would be to make the argument accept only a Buffer, and force the caller to decide how they want to convert.  You could provide a `passwordToBuffer` helper if you want to help make that use case clearer.  `bigNumberishToBuffer` already exists for the other cases.  A less strict alternative would be to accept `Buffer | string` (but not `bigint`), though there's still room for a little bit of confusion if the caller has a bigint in a hex or decimal string.  I could also imagine having two different versions of this function: e.g. `signMessageWithPassword` and `signMessageWithKey`, though I think that's more awkward than the extra helper function.

2) Unrelated to this private key, but related to my note about truncation: If you agree with my assumption that Bignumberish is intended to always represent field element, it might make sense to make that clear in the docs and/or the name.  In that case it would also make sense for the buffer conversion to always use a size of 32 bytes, and to pad with 0s at the front (big-endian) or end (little-endian).  It would also make sense to check size on user-provided buffers where it's not being checked already.

Let me know what you think of these suggestions.  I'm guessing at a lot of context here, so they may not be perfect, but they're the best I've got right now.

0xjei (approved):

packages/eddsa-poseidon/src/eddsa-poseidon.ts
0xjei: Rushing into this conversation now. I appreciate the detailed feedback and insights shared regarding the handling of flexible inputs, especially in the context of private keys and the potential confusion it introduces. Your experience @artwyman underscores the critical balance between flexibility and clarity. My two cents:

- Restricting the argument types for private keys to `Buffer` is a giant step for eliminating ambiguity and aligns fine with `zk-kit` ethos of leaving more freedom while making the users decide what kind of pre-conversion process is needed before passing the input to the method.
- Standardizing the buffer conversion process and adding more documentation is a great way to maintain consistency and reliability across the whole toolkit, so +1.

packages/utils/src/error-handlers.ts
```diff
@@ -0,0 +1,168 @@
...
+} from "./type-checks"
+
+/**
+ * It throws a type error if the parameter value has not been defined.
+ * @param parameterValue The parameter value.
+ * @param parameterName The parameter name.
+ */
+export function requireDefined(parameterValue: any, parameterName: string) {
+    if (!isDefined(parameterValue)) {
+        throw new TypeError(`Parameter '${parameterName}' is not defined`)
```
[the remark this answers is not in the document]
cedoor: Including the value might be tricky because it could be an object without any `.toString()` method, or an array, or any type that would make the message hard to read.

I've added the type to the error message where the check is on a native type. Having:

`Parameter '${parameterName}' is not a stringified bigint, but received type: string` could not be very precise 🤔

packages/utils/src/error-handlers.ts
```diff
@@ -0,0 +1,168 @@
...
+ * @param parameterName The parameter name.
+ */
+export function requireArray(parameterValue: any[], parameterName: string) {
+    if (!isArray(parameterValue)) {
+        throw new TypeError(`Parameter '${parameterName}' is not an array`)
+    }
+}
+
+/**
+ * It throws a type error if the parameter value is not a uint8array.
```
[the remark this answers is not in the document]
cedoor: I thought of using it in lowercase to follow the way the `bigint` type is handled. There is the class `BigInt` and the type `bigint`. Let me know if it makes sense for you.

packages/utils/src/conversions.ts
```diff
@@ -161,13 +161,8 @@ export function bigIntToBuffer(n: bigint): Buffer {
...
-        typeof n === "bigint" ||
-        (typeof n === "string" && isStringifiedBigint(n)) ||
-        (typeof n === "string" && isHexadecimal(n))
-    ) {
-        return BigInt(n)
+    if (isNumber(n) || isBigInt(n) || isStringifiedBigInt(n) || isHexadecimal(n)) {
+        return BigInt(n as BigNumber | number)
     }
 
     return bufferToBigInt(n as Buffer)
```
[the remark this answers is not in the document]
cedoor: Oh, good catch thanks! Could you open another issue for this?

packages/eddsa-poseidon/src/eddsa-poseidon.ts
cedoor: Thanks guys!

I think using Buffer only (and Uint8Array maybe) for the reasons you explained sounds right in this case. My only doubt is that Buffer could not be the most common type used here, as it is not the best type to export for storing private keys. Devs usually use hexadecimal strings and they'd need to convert it every time they want to sign a message. We could provide additional utility functions, but it's worth considering the additional work they have to do in relation to the ambiguity that having multiple types entails. That said, it also depends on the context imo, and in this case, as it is an implementation of a specific algorithm, perhaps it makes sense to be more restrictive.

> If you agree with my assumption that Bignumberish is intended to always represent field element.

I think most Bignumberish inputs are 32-byte values in this library, but that could not be the case in other apps/libraries. What about adding an optional `size` parameter to the bigNumberish functions as well? So that `eddsa-poseidon` can define it.

cedoor: Also, we could create separate issues for these conversations and close this PR. It looks like there are three:

1. Restrict the number of supported types for the EdDSA `privateKey` parameter
2. Add `size` parameter to the bigNumberish functions - ish
3. Check the length of arbitrary strings for the EdDSA `message` parameter

Other issues could be included in these 3. Missing anything?

packages/utils/src/error-handlers.ts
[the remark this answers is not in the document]
artwyman: I've noticed that for `bigint`.  For `Uint8array` I've only seen this spelling, which is a bit suprising since I'd have assumed `UInt8Array` with a capital I.  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array

JavaScript might be making a distinction between primitive types (lowercase - `bigint`) vs. object types (uppercase - `BigInt`) but I'm not familiar enough to be sure.

I don't really have a strong opinion on this issue.  I'm just pointing out an inconsistency which stuck out to me.

packages/eddsa-poseidon/src/eddsa-poseidon.ts
artwyman: I'm good with breaking things out.  Quick responses to your latest thoughts:

Size parameters seem like a lot of overhead, particularly if they're usually forced to be 32.  Using a type like Buffer/Uint8Array/string which has its own size seems like the cleanest way to do that, but I think bigint is actually the clearest/simplest type for most of this library since it avoids byte order issues.  As discussed elsewhere, this private key is distinct because it's an input of bytes, not of a field element.

I think the question of which input encoding (Buffer, hex string, etc) for private keys is fuzzy enough that it's going to be hard to do right for all users.  In Zupass, I've standardized on hex strings, and we have `toHexString`/`fromHexString` helpers to convert from/to Buffers.  That's mostly because that was the standard set by the EdDSATicketPCD, and I want simplicity and clarity.  There are situations in which I might consider performance most important, where I might already have my bytes in a Buffer and resent the need to convert it to hex unnecessarily.  But I come from lower-level languages than JavaScript, so I'm not sure how much that level of micro-optimization really matters.

artwyman (approved):
As discussed, I think this PR can go ahead, with the understanding that some of the ongoing discussions will be continued in new Issues/PRs.

packages/eddsa-poseidon/src/eddsa-poseidon.ts
artwyman: Just to add another wrinkle to string encodings: The Zupass/PCD standard is hex strings without the 0x prefix.  I think that actually makes sense when you're encoding an arbitrary-length array of bytes, because 0x to me suggests a single number.  But it is part of what lead to the ambiguity which started this conversation (because my test key happened to contain only the digits 0-9).

packages/utils/src/error-handlers.ts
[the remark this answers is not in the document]
cedoor: > JavaScript might be making a distinction between primitive types (lowercase - `bigint`) vs. object types (uppercase - `BigInt`) but I'm not familiar enough to be sure.
 
There isn't really a primitive type in this case, so it might make sense to change a bit doc and names to reflect that we are actually checking an object type maybe?

I think that makes sense. Let me work on it to wrap this PR up.

packages/eddsa-poseidon/src/eddsa-poseidon.ts
cedoor: >  There are situations in which I might consider performance most important.

That's the point I think. In this case, we have an algorithm implementation and even if it is JS perhaps it is better to minimize conversions and type-checks in functions and prioritize performances, but also give devs an optimal experience. I created an issue where we can continue our discussion: https://github.com/privacy-scaling-explorations/zk-kit/issues/188

cedoor: > I think that actually makes sense when you're encoding an arbitrary-length array of bytes.

That makes sense. If we add new utility functions to convert hex to buffer or buffer to hex we shouldn't use `0x`. In the other cases, it makes sense to use `0x` as they are mostly conversions between bigints and hexadecimals.

0xjei (approved):

PR #184, opened 2024-02-28
More documentation and tests on `@zk-kit/eddsa-poseidon` package

### LINKED ISSUE #140: Add tests and documentation to `@zk-kit/eddsa-poseidon`

0xjei:
## Description
This PR adds more documentation to the `eddsa-poseidon` package, with the aim of enhancing the code readability with meaningful information.

## Related Issue(s)
closes #140 

## Other information

This PR adds the missing tests to reach the 100% code coverage.

## Checklist

### DIFF
--- a/packages/eddsa-poseidon/src/blake.ts
+++ b/packages/eddsa-poseidon/src/blake.ts
@@ -5,7 +5,7 @@ import { Blake512 } from "blake-hash/lib"
  * Computes the Blake512 hash of the input message.
  * Blake512 is a cryptographic hash function that produces a hash value of 512 bits,
  * commonly used for data integrity checks and other cryptographic applications.
- * @param message - The input data to hash, provided as a Buffer.
+ * @param message The input data to hash, provided as a Buffer.
  * @returns A Buffer containing the 512-bit hash result.
  */
 export default function hash(message: Buffer): Buffer {
--- a/packages/eddsa-poseidon/src/blake.ts
+++ b/packages/eddsa-poseidon/src/blake.ts
@@ -1,6 +1,13 @@
 // @ts-ignore
 import { Blake512 } from "blake-hash/lib"
 
+/**
+ * Computes the Blake512 hash of the input message.
+ * Blake512 is a cryptographic hash function that produces a hash value of 512 bits,
+ * commonly used for data integrity checks and other cryptographic applications.
+ * @param message - The input data to hash, provided as a Buffer.
+ * @returns A Buffer containing the 512-bit hash result.
+ */
 export default function hash(message: Buffer): Buffer {
     const engine = new Blake512()
 
--- a/packages/eddsa-poseidon/src/eddsa-poseidon.ts
+++ b/packages/eddsa-poseidon/src/eddsa-poseidon.ts
@@ -25,14 +25,15 @@ import { Signature } from "./types"
 import * as utils from "./utils"
 
 /**
- * Hashes the 32-byte private key using Blake1, prunes the lower 32 bytes
- * of the buffer and converts it to a little-endian integer.
- * This function is used to obtain the secret scalar to be used as
- * the input of the private key in the circuits, since the circuit only
- * performs a fixed-base scalar multiplication.
- * For more info about these steps: {@link https://datatracker.ietf.org/doc/html/rfc8032#section-5.1.5}.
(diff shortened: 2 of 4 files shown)


packages/eddsa-poseidon/src/blake.ts
```diff
@@ -1,6 +1,13 @@
 // @ts-ignore
 import { Blake512 } from "blake-hash/lib"
 
+/**
+ * Computes the Blake512 hash of the input message.
+ * Blake512 is a cryptographic hash function that produces a hash value of 512 bits,
+ * commonly used for data integrity checks and other cryptographic applications.
+ * @param message - The input data to hash, provided as a Buffer.
```
cedoor, line 8:
```suggestion
 * @param message The input data to hash, provided as a Buffer.
```

We're not using the hyphen in the other packages. We should keep it consistent imo.

packages/eddsa-poseidon/src/eddsa-poseidon.ts
```diff
@@ -177,23 +189,46 @@ export function unpackPublicKey(publicKey: BigNumber): Point<string> {
...
 }
 
+/**
+ * Represents a cryptographic entity capable of signing messages and verifying signatures
+ * using the EdDSA scheme with Poseidon hash and the Baby Jubjub elliptic curve.
+ */
 export class EdDSAPoseidon {
+    // Private key for signing, stored securely.
     privateKey: BigNumberish
+    // The secret scalar derived from the private key.
```
cedoor, line 199: Worth adding some comments about why this attribute is defined.

cedoor (changes requested):

packages/eddsa-poseidon/src/eddsa-poseidon.ts
0xjei: "The secret scalar derived from the private key __to compute the public key__".

This should be sufficient, as more information can be retrieved using the `deriveSecretScalar()` function. What do you think @cedoor?

0xjei:
```ts
/**
 * Derives a secret scalar from a given EdDSA private key.
 * This process involves hashing the private key with Blake1, pruning the resulting hash to retain the lower 32 bytes,
 * and converting it into a little-endian integer. The use of the secret scalar streamlines the public key generation
 * process by omitting steps 1, 2, and 3 as outlined in RFC 8032 section 5.1.5, enhancing circuit efficiency and simplicity.
 * This method is crucial for fixed-base scalar multiplication operations within the correspondent cryptographic circuit.
 * For detailed steps, see: {@link https://datatracker.ietf.org/doc/html/rfc8032#section-5.1.5}.
 * @param privateKey The EdDSA private key for generating the associated public key.
 * @returns The derived secret scalar, optimized for circuit calculations.
 */
 ```
 Do you think is worth linking the circuit as an example of implementation?

File: packages/eddsa-poseidon/tests/index.test.ts [hunk 3]
```diff
@@ -227,6 +238,20 @@ describe("EdDSAPoseidon", () => {
         expect(unpackedPublicKey[1]).toBe(publicKey[1])
     })
 
+    it("Should not unpack a public key if the public key is not on the curve", async () => {
+        const fun = () => unpackPublicKey("e")
+
+        expect(fun).toThrow("Invalid public key")
```
cedoor, line 244: Isn't this a type error?

cedoor: @0xjei If it's a type error, the test message is not the right one.

0xjei: What kind of message do you prefer?

cedoor: If it's a type error, I think you should either specify that it's throwing a type error or changing the input to meet the test message.

0xjei: Hmm the current message is "Should not unpack a public key if the public key type is not supported" so I mentioning the type. I think you're seeing an outdated view

cedoor: Yes, I saw the old message! It's ok then 👍🏽

cedoor (approved):

PR #193, opened 2024-03-07
Improve `@zk-kit/utils` package

### LINKED ISSUE #141: Add tests and documentation to `@zk-kit/utils`

### LINKED ISSUE #143: Move utils functions of `@zk-kit/circuits` and `@zk-kit/poseidon-cipher` to `@zk-kit/utils`

### LINKED ISSUE #144: Make `babyjubjub`/`utils` packages browser-compatible

0xjei:
## Description
This PR aims to make the `@zk-kit/utils` package and all packages that use it, such as `@zk-kit/baby-jubjub`, browser compatible. To achieve this, the [buffer](https://www.npmjs.com/package/buffer) cross-compatible (node / browser) package will be used, and a `getRandomValue()` utility will be used to replace the now-deprecated `genRandomBabyJubJubValue()`. In addition, documentation and tests are added to maximise coverage.

## Related Issue(s)
closes #141 
closes #143  
closes #144 

## Checklist


packages/baby-jubjub/rollup.iife.config.ts
```diff
@@ -35,13 +33,5 @@ export default {
...
-    plugins: [
-        typescript({ tsconfig: "./build.tsconfig.json" }),
-        commonjs(),
-        nodeResolve({
-            preferBuiltins: true
-        }),
-        nodePolyfills({ include: null }),
-        cleanup({ comments: "jsdoc" })
-    ]
+    plugins: [typescript({ tsconfig: "./build.tsconfig.json" }), commonjs(), cleanup({ comments: "jsdoc" })]
```
cedoor, line 36: It shouldn't need `commonjs`, and it can be merged with the `rollup.config.ts` I think.

packages/utils/src/crypto/crypto.node.ts
```diff
@@ -0,0 +1,13 @@
+import { randomBytes as rb } from "crypto"
```
cedoor, line 1: It shouldn't need the alias here.

packages/utils/src/crypto/crypto.browser.ts
```diff
@@ -0,0 +1,13 @@
+import { getRandomValues } from "crypto"
```
cedoor, line 1: There shouldn't be any import here. `crypto` is still node.

You should be able to use `crypto.getRandomValues` below, without any import on browsers.

packages/utils/src/crypto/crypto.browser.ts
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * Generates a random bigint value. It uses the Web Cryptography API to securely generate
+ * a random sequence of size bytes and then converts those bytes into a bigint using
+ * big-endian byte order.
+ * @param size The number of bytes.
+ * @returns A randomly generated bigint value.
+ */
+/* eslint-disable import/prefer-default-export */
+export function getRandomValue(size: number): bigint {
```
cedoor, line 11: I think the two `getRandomValue` should both return a `Uint8Array` (i.e. a buffer) here. They can convert that buffer to a bigint or whatever with the `conversions` module's functions.

`randomBytes` (on Node) and `getRandomValues` (on browsers) return a buffer, which could be more than 1 value/number, so using the plural sounds more correct here. This is another reason why converting it to bigint here could lead to some issues with its size (bigints don't have the same limit as buffers in size).

packages/utils/src/index.ts
```diff
@@ -4,8 +4,10 @@ import * as conversions from "./conversions"
 import * as packing from "./proof-packing"
 import * as typeChecks from "./type-checks"
 import * as errorHandlers from "./error-handlers"
+import * as cryptoNode from "./crypto/crypto.node"
+import * as cryptoBrowser from "./crypto/crypto.browser"
```
cedoor, line 8:
```suggestion
import * as crypto from "./crypto/crypto.node"
```

There should be 1 import only here. The Rollup alias plugin replaces "./crypto/crypto.node" with "./crypto/crypto.browser" during the building process.

For example: https://github.com/semaphore-protocol/semaphore/blob/43f555e229fd25b02b3cacf0d1a9ca0b39b68c4a/packages/identity/src/identity.ts#L5

packages/circuits/tests/poseidon-cipher.test.ts
```diff
@@ -1,13 +1,13 @@
 import { WitnessTester } from "circomkit"
 import { Nonce, PlainText, poseidonDecrypt, poseidonEncrypt, poseidonPerm } from "@zk-kit/poseidon-cipher"
-
-import { circomkit, genEcdhSharedKey, genPublicKey, genRandomBabyJubValue } from "./common"
+import { cryptoNode } from "@zk-kit/utils"
```
cedoor, line 3: It would better to use selective imports when possible, i.e. "@zk-kit/utils/crypto"

packages/eddsa-poseidon/rollup.iife.config.ts
```diff
@@ -35,13 +33,5 @@ export default {
...
-    plugins: [
-        typescript({ tsconfig: "./build.tsconfig.json" }),
-        commonjs(),
-        nodeResolve({
-            preferBuiltins: true
-        }),
-        nodePolyfills({ include: null }),
-        cleanup({ comments: "jsdoc" })
-    ]
+    plugins: [typescript({ tsconfig: "./build.tsconfig.json" }), commonjs(), cleanup({ comments: "jsdoc" })]
```
cedoor, line 36: `@zk-kit/eddsa-poseidon` includes `blake-hash` in its bundle, which uses the NodeJS Buffer class. If we want to remove these Rollup plugins, we should copy the `blake-hash` code we need and replace Buffer with the npm `buffer` package. Just a possible proposal.

packages/utils/rollup.node.config.ts
```diff
@@ -1,6 +1,7 @@
 import typescript from "@rollup/plugin-typescript"
 import fs from "fs"
 import cleanup from "rollup-plugin-cleanup"
+import json from "@rollup/plugin-json"
```
cedoor, line 4: Why do we need this plugin?

packages/utils/rollup.node.config.ts
```diff
@@ -16,10 +17,17 @@ export default [
     {
         input: "src/index.ts",
         output: [
-            { file: pkg.exports["."].require, format: "cjs", banner },
-            { file: pkg.exports["."].default, format: "es", banner }
+            { file: pkg.exports["."].node.require, format: "cjs", banner, exports: "auto" },
+            { file: pkg.exports["."].node.default, format: "es", banner }
         ],
-        plugins: [typescript({ tsconfig: "./build.tsconfig.json" }), cleanup({ comments: "jsdoc" })]
+        external: [...Object.keys(pkg.dependencies)],
```
cedoor, line 23:
```suggestion
external: Object.keys(pkg.dependencies),
```

packages/utils/package.json
```diff
@@ -44,6 +48,16 @@
...
+        },
+        "./crypto.browser": {
+            "types": "./dist/types/crypto.browser.d.ts",
+            "require": "./dist/lib.commonjs/crypto.browser.cjs",
+            "default": "./dist/lib.esm/crypto.browser.js"
+        },
+        "./crypto.node": {
+            "types": "./dist/types/crypto.node.d.ts",
+            "require": "./dist/lib.commonjs/crypto.node.cjs",
+            "default": "./dist/lib.esm/crypto.node.js"
```
cedoor, line 60: This should be something like:
```json
"./crypto": {
    "types": "./dist/types/crypto.browser.d.ts",
    "node": {
      "types": "./dist/types/crypto.node.d.ts",  
      "require": "./dist/lib.commonjs/crypto.node.cjs",
      "default": "./dist/lib.esm/crypto.node.js"
    },
    "browser": "./dist/lib.esm/crypto.browser.js",
    "default": "./dist/lib.esm/crypto.browser.js"
}
```

packages/utils/package.json
```diff
@@ -7,13 +7,17 @@
     "iife": "dist/index.iife.js",
     "unpkg": "dist/index.iife.min.js",
     "jsdelivr": "dist/index.iife.min.js",
-    "main": "dist/index.js",
+    "main": "dist/index.node.js",
```
cedoor, line 10:
```suggestion
    "main": "dist/index.browser.js",
```

packages/utils/package.json
```diff
@@ -7,13 +7,17 @@
...
         ".": {
             "types": "./dist/types/index.d.ts",
-            "require": "./dist/index.cjs",
-            "default": "./dist/index.js"
+            "node": {
+                "require": "./dist/index.node.cjs",
+                "default": "./dist/index.node.js"
+            },
+            "browser": "./dist/index.browser.js",
+            "default": "./dist/index.node.js"
```
cedoor, line 20:
```suggestion
            "default": "./dist/index.browser.js"
```

cedoor (changes requested):

packages/utils/src/crypto/crypto.browser.ts
0xjei: my bad, you are right!

packages/utils/src/crypto/crypto.browser.ts
0xjei: Thank you so much. Therefore, I will change `bigint` with `Uint8Array` and use the correspondent converters as needed.

packages/eddsa-poseidon/package.json
```diff
@@ -40,7 +40,7 @@
         "@rollup/plugin-node-resolve": "^15.2.3",
         "@rollup/plugin-terser": "^0.4.4",
         "@rollup/plugin-typescript": "^11.1.6",
-        "blake-hash": "2.0.0",
+        "buffer": "6.0.3",
```
cedoor, line 43: It should be a dev dependency.

cedoor (changes requested):

packages/circuits/package.json
```diff
@@ -26,6 +26,7 @@
         "@zk-kit/eddsa-poseidon": "0.5.1",
         "@zk-kit/poseidon-cipher": "0.1.1",
         "@zk-kit/smt": "1.0.0",
+        "@zk-kit/utils": "0.4.3",
```
cedoor, line 29: This needs to be updated to 0.5.0

packages/eddsa-poseidon/package.json
```diff
@@ -51,6 +50,7 @@
     },
     "dependencies": {
         "@zk-kit/baby-jubjub": "0.2.0",
-        "@zk-kit/utils": "0.4.3"
+        "@zk-kit/utils": "0.4.3",
```
cedoor, line 53: Need to update it to 0.5.0

packages/eddsa-poseidon/tests/index.test.ts
```diff
@@ -197,11 +198,12 @@ describe("EdDSAPoseidon", () => {
...
         for (let i = 0, len = 10; i < len; i += 1) {
-            const privateKey = crypto.randomBytes(32)
+            const privateKey = crypto.getRandomValues(32)
+            const bgPrivateKey = beBufferToBigInt(Buffer.from(privateKey))
 
-            const publicKey = derivePublicKey(privateKey)
+            const publicKey = derivePublicKey(bgPrivateKey)
 
-            const circomlibPublicKey = eddsa.prv2pub(privateKey)
+            const circomlibPublicKey = eddsa.prv2pub(beBigIntToBuffer(bgPrivateKey))
```
cedoor, line 206:
```ts
const privateKey = crypto.getRandomValues(32)
const publicKey = derivePublicKey(privateKey)
const circomlibPublicKey = eddsa.prv2pub(privateKey)
```

Should this still work?

packages/utils/package.json
```diff
@@ -44,6 +48,21 @@
             "types": "./dist/types/f1-field.d.ts",
             "require": "./dist/lib.commonjs/f1-field.cjs",
             "default": "./dist/lib.esm/f1-field.js"
+        },
+        "./blake512": {
+            "types": "./dist/types/blake512.d.ts",
+            "require": "./dist/lib.commonjs/blake512.cjs",
+            "default": "./dist/lib.esm/blake512.js"
+        },
```
cedoor, line 56: We should remove this. `blake512` is part of `eddsa-poseidon` now.

packages/utils/src/crypto/crypto.browser.ts
```diff
@@ -0,0 +1,15 @@
...
+ */
+/* eslint-disable import/prefer-default-export */
+export function getRandomValues(size: number): Uint8Array {
+    if (size <= 0) throw Error(`size ${size} is too small, need at least 1`)
+
+    const byteArray = new Uint8Array(size)
+
+    crypto.getRandomValues(byteArray)
+
+    return byteArray
```
cedoor, line 14:
```suggestion
    return crypto.getRandomValues(new Uint8Array(size))
```

Possible improvement

packages/utils/src/scalar.ts
```diff
@@ -1,23 +1,73 @@
+/**
+ * @module ScalarOperations
```
cedoor, line 2: Could be more consistent to use the file name here, like the other modules.

packages/utils/tests/crypto.test.ts
```diff
@@ -0,0 +1,14 @@
+import { crypto } from "../src"
+
+describe("crypto", () => {
+    it("Should throw when size is zero", async () => {
+        const fun = () => crypto.getRandomValues(0)
+
+        expect(fun).toThrow("size 0 is too small, need at least 1")
+    })
+    it("Should correctly return a random value", async () => {
+        expect(crypto.getRandomValues(1)).toBeDefined()
```
cedoor, line 10: It could also check the length

packages/utils/src/scalar.ts
```diff
@@ -1,23 +1,73 @@
...
+ * This module provides utility functions for performing scalar operations
+ * within a finite field, especially designed to handle operations on bigints.
```
cedoor, line 4: There doesn't have to be a finite field here I think, not necessarily

cedoor (changes requested):

packages/eddsa-poseidon/tests/index.test.ts
0xjei: Nope, you'll get an error on the `derivePublicKey(privateKey)` call.

```bash
Argument of type 'Uint8Array' is not assignable to parameter of type 'BigNumberish'.
  Type 'Uint8Array' is missing the following properties from type 'Buffer': write, toJSON, equals, compare, and 66 more.ts(2345)
```

cedoor:
```ts
const privateKey = crypto.getRandomValues(32)
const publicKey = derivePublicKey(Buffer.from(privateKey))
const circomlibPublicKey = eddsa.prv2pub(privateKey)
```

? Just to avoid converting buffer to bigint and vice versa.

0xjei:
```ts
            const privateKey = crypto.getRandomValues(32)
            const publicKey = derivePublicKey(Buffer.from(privateKey))
            const circomlibPublicKey = eddsa.prv2pub(Buffer.from(privateKey))
```

This is correct

cedoor: Great, then also:
```ts
            const privateKey = Buffer.from(crypto.getRandomValues(32))
            const publicKey = derivePublicKey(privateKey)
            const circomlibPublicKey = eddsa.prv2pub(privateKey)
```

cedoor (approved):
Great work <3

PR #194, opened 2024-03-07
Add new utility functions to convert hex and buffers

### LINKED ISSUE #189: New utility functions to convert hexadecimals to buffers and vice versa

There is currently no utility function for converting bytes to hexadecimals and vice versa, which could be useful when one wants to use a function that only accepts bytes or hexadecimals. In this case, hexadecimal makes sense not to start with `0x`, since buffers are also (and probably mostly) used for arbitrary-length bytes, and not for single numbers.

If using `Buffer` (NodeJS class or `buffer` NPM package), the following code should work:

```ts
export function hexadecimalToBuffer(value: string): Buffer {
    return Buffer.from(value, "hex")
}

export function bufferToHexadecimal(value: Buffer | Uint8Array): string {
    return Buffer.from(value).toString("hex")
}
```

**Additional context**

Originated from discussion in https://github.com/privacy-scaling-explorations/zk-kit/pull/178#discussion_r1513494373.

cedoor:
## Description

This PR adds 4 functions to the `conversions` module to convert hexadecimal strings to buffers and vice versa, both using little-endian and big-endian byte orders.

It also adds `isBuffer` to the `type-checks` module, and a `requireBuffer` to the `error-handlers` module.

Finally, it adds a new `prefix` parameter to the `isHexadecimal` and `requireHexadecimal` functions. It is an optional boolean (set to `true` by default) to consider or not the `0x` or `0X` prefix in a hexadecimal string.

## Related Issue(s)

Closes #189

## Checklist

### DIFF
--- a/packages/utils/src/conversions.ts
+++ b/packages/utils/src/conversions.ts
@@ -186,70 +186,24 @@ export function bigNumberishToBuffer(n: BigNumberish): Buffer {
 
 /**
  * Converts an hexadecimal string to a buffer. The hexadecimal string
- * should not start with '0x' or '0X'.
- * It uses a big-endian byte order.
+ * should not start with '0x' or '0X'. It keeps the bytes in the same order.
  * @param value The hexadecimal string to convert.
  * @returns The buffer representation of the hexadecimal string.
  */
-export function beHexadecimalToBuffer(value: string): Buffer {
+export function hexadecimalToBuffer(value: string): Buffer {
     requireHexadecimal(value, "value", false)
 
     return Buffer.from(value, "hex")
 }
 
 /**
- * Converts an hexadecimal string to a buffer. The hexadecimal string
- * should not start with '0x' or '0X'.
- * It uses a little-endian byte order.
- * @param value The hexadecimal string to convert.
- * @returns The buffer representation of the hexadecimal string.
- */
-export function leHexadecimalToBuffer(value: string): Buffer {
-    requireHexadecimal(value, "value", false)
-
-    return Buffer.from(value, "hex").reverse()
-}
-
-/**
- * Converts an hexadecimal string to a buffer. Alias for beHexadecimalToBuffer.
- * @param value The hexadecimal string to convert.
- * @returns The buffer representation of the hexadecimal string.
- */
-export function hexadecimalToBuffer(value: string): Buffer {
-    return beHexadecimalToBuffer(value)
-}
-
-/**
  * Converts a buffer to a hexadecimal string. It accepts 'Buffer' or 'Uint8Array'.
- * The hexadecimal string will not start with '0x' or '0X'.
- * It uses a big-endian byte order.
- * @param value The buffer to convert.
- * @returns The converted hexadecimal string.
- */
-export function beBufferToHexadecimal(value: Buffer | Uint8Array): string {
-    requireTypes(value, "value", ["Buffer", "Uint8Array"])
-
-    return Buffer.from(value).toString("hex")
-}
-
-/**
- * Converts a buffer to a hexadecimal string. It accepts 'Buffer' or 'Uint8Array'.
- * The hexadecimal string will not start with '0x' or '0X'.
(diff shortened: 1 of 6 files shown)


0xjei (approved):

packages/utils/src/type-checks.ts
```diff
@@ -111,11 +120,17 @@ export function isStringifiedBigInt(value: any): boolean {
 
 /**
  * Checks if a string is a valid hexadecimal string representation.
- * The string must start with '0x' or '0X' followed by one or more hexadecimal digits (0-9, a-f, A-F).
+ * If 'prefix' is not defined and set to 'false', the string must start with '0x' or '0X'
```
artwyman, line 123: I think this explanation has it backwards?  prefix=true means the string must start with the prefix.
You should also state explicitly that when prefix is false or undefined, the string is required to not have a prefix.  I.e. the prefix is never optional.

packages/utils/src/conversions.ts
```diff
@@ -176,9 +177,79 @@ export function bigNumberishToBigInt(n: BigNumberish): bigint {
...
+        return n as Buffer
     }
 
     return bigIntToBuffer(bigNumberishToBigInt(n))
 }
+
+/**
+ * Converts an hexadecimal string to a buffer. The hexadecimal string
+ * should not start with '0x' or '0X'.
+ * It uses a big-endian byte order.
```
artwyman, line 190: I'm not sure that endian-ness is meaningful on byte order kind of conversion.  There's no "most significant byte" in either the input or the output, just a sequence of bytes indexed 0...N.  Byte order comes up when converting bytes into numbers which are bigger than a byte.

I'd suggest you only have one version of this conversion (the one without the reverse) and simply explain that it keeps the bytes in the same order, after grouping the hex digits in 2-digit chunks.

Technically (but not relevant and not worth changing or documenting) the conversion of 2 hex digits to a byte has a digit order (not a byte order) which is big-endian, but as with normal hex and decimal, I think that's obvious enough to be assumed.  Digits always put the most-significant digit first.  You'd never write 5*2 as 01 rather than 10.

packages/utils/src/type-checks.ts
cedoor: > I think this explanation has it backwards? prefix=true means the string must start with the prefix.

Yes, my bad!

> You should also state explicitly that when prefix is false or undefined, the string is required to not have a prefix. I.e. the prefix is never optional.

Well it is optional, isn't it? Devs could decide to not pass any parameter and it will be `true` by default. Also, it will not undefined unless devs set it to `undefined` explicitly.

packages/utils/src/conversions.ts
cedoor: Oh this byte-order stuff still confuses me a bit 😅  Thanks for this explanation!

packages/utils/src/type-checks.ts
artwyman: Sorry, I wasn't clear, but you understood my intent.  The prefix argument is optional.  What I meant is that the `0x` prefix is not optional.  It's either required, or it's not allowed, depending on the argument value.  My one suggestion to clarify that in the latest wording would be to change "otherwise no prefix is needed." to "otherwise no prefix is expected." or "allowed" to make it clear that the parser won't accept a prefix in that mode.

packages/utils/src/conversions.ts
artwyman: It's super unclear, and I probably don't have the terminology perfect either.  Thanks for putting up with my pedantry.

artwyman (approved):

packages/utils/src/conversions.ts
cedoor: Oh no, please be pedantic. It's what a PR needs 👍🏽  Thanks!

packages/utils/src/type-checks.ts
cedoor: Sorry, needed was not the right term indeed! Thanks

PR #196, opened 2024-03-13
Build libraries with proper Rollup configuration

cedoor:
## Description

This PR fixes some Rollup configurations that were not working properly, adjusting PR #193 output before publishing new versions.

## Related Issue(s)

Re #144

## Checklist


packages/baby-jubjub/rollup.config.ts
```diff
@@ -15,24 +17,45 @@ const banner = `/**
...
+                plugins: [terser({ output: { preamble: banner } })]
+            }
+        ],
+        external: [],
+        plugins: [
+            typescript({
+                tsconfig: "./build.tsconfig.json"
+            }),
+            nodeResolve({ preferBuiltins: false, browser: true }),
+            nodePolyfills(),
```
0xjei, line 57: Why do we need the polyfills here? Is this a problem related to the `buffer`?

packages/eddsa-poseidon/rollup.config.ts
```diff
@@ -14,19 +16,53 @@ const banner = `/**
...
+const name = pkg.name.split("/")[1].replace(/[-/]./g, (x: string) => x.toUpperCase()[1])
+
+export default [
+    {
+        input: "src/index.ts",
+        output: [
+            { file: pkg.exports["."].require, format: "cjs", banner },
+            { file: pkg.exports["."].default, format: "es", banner }
+        ],
+        external: [
```
0xjei, line 28: Why are we putting them as external?

packages/utils/package.json
```diff
@@ -48,16 +48,6 @@
             "types": "./dist/types/f1-field.d.ts",
             "require": "./dist/lib.commonjs/f1-field.cjs",
             "default": "./dist/lib.esm/f1-field.js"
-        },
-        "./crypto": {
```
0xjei, line 52: why are we removing this?

0xjei (approved):
Works fine, thank you for improving the build process. Just left some comments to better understand the configs but ready to merge

packages/baby-jubjub/rollup.config.ts
cedoor: Yes, it looks like Rollup still wants the polyfill for `buffer`, even if we're using a package (from which polyfill is based I think). I've tried with `commonjs` and `node-resolve` but they're not enough.

packages/eddsa-poseidon/rollup.config.ts
cedoor: Because we don't want them to be included in the final bundle. They should be external dependencies.

packages/utils/package.json
cedoor: The Rollup conf with `preserveModules: true`, which is used to compile the ts files to js without bundling in 1 file, doesn't include the `crypto/crypto.browser.ts` file, because it is not actually imported by `index.ts`. I haven't found any simple solution so far, so I opted not to use that conditional export for now. `crypto` can be imported from `@zk-kit/utils` only for new.

PR #200, opened 2024-03-14
fix(eddsa-poseidon): eddsa-poseidon should accept Points and Signatures in numeric form

### LINKED ISSUE #198: eddsa-poseidon rejects Points and Signatures in numeric form
**Describe the bug**
The verifySignature and unpackPublicKey are declared to accept inputs of type `Point` and `Signature` (using default type parameters.  This means `Point<BigNumber>` and `Signature<BigNumber>` so all numbers should be able to be `string` or `bigint`.

**To Reproduce**

Modify a unit test or other code to pass an input of type Point<bigint> or Signature<bigint>.

**Expected behavior**

Numeric inputs should be allowed to already be bigint, without requiring string conversion.

**Additional context**

Discovered in zupass repo when updating to the latest version.
This apparently broke since an earlier version of the zk-kit packages, since my code used to work.
Here's a PR containing the upgrade, and the workaround: https://github.com/proofcarryingdata/zupass/pull/1588

artwyman:
## Description

Added `isBigNumber` and `requireBigNumber` in zk-kit/utils
Used `isBigNumber` in `isPoint` and `isSignature` in zk-kit/eddsa-poseidon to make them more accepting

## Related Issue(s)

Fixes #198

### DIFF
--- a/packages/utils/src/type-checks.ts
+++ b/packages/utils/src/type-checks.ts
@@ -143,7 +143,7 @@ export function isHexadecimal(value: any, prefix = true) {
  * @param value The value to check.
  */
 export function isBigNumber(value: any): boolean {
-    return isBigInt(value) || isStringifiedBigInt(value);
+    return isBigInt(value) || isStringifiedBigInt(value)
 }
 
 /**
--- a/packages/utils/src/type-checks.ts
+++ b/packages/utils/src/type-checks.ts
@@ -21,6 +21,7 @@ const supportedTypes = [
     "bigint",
     "stringified-bigint",
     "hexadecimal",
+    "bignumber",
     "bignumberish"
 ] as const
 
@@ -185,6 +186,8 @@ export function isType(value: any, type: SupportedType): boolean {
             return isStringifiedBigInt(value)
         case "hexadecimal":
             return isHexadecimal(value)
+        case "bignumber":
+            return isBigNumber(value)
         case "bignumberish":
             return isBigNumberish(value)
         default:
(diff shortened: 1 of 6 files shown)


packages/utils/src/type-checks.ts
```diff
@@ -136,6 +136,16 @@ export function isHexadecimal(value: any, prefix = true) {
     return /^[0-9a-fA-F]+$/.test(value)
 }
 
+/**
+ * Checks if the given value can be considered as BigNumber.
+ * A value is considered a BigNumber if it is a bigint or a string
+ * that can be converted to a bigint (via `Bigint(s)`).
+ * @param value The value to check.
+ */
+export function isBigNumber(value: any): boolean {
```
cedoor, line 145: Could you also add `big-number` to the `supportedType` list and to the `isType` function (and add tests for it)?

I also wonder if `big-number` and `bigint` could be confusing. What do you think?

cedoor (changes requested):

packages/utils/src/type-checks.ts
artwyman: There's definitely the potential for confusion.  I recently found myself misinterpreting `BigNumber` as being a big version of `number` and worried that it might be a non-integer.

That said, this PR isn't changing the situation.  The types `BigNumber` and `BigNumberish` defined in zk-kit/util introduce that terminology.  I think it's probably not a big issue so long as the usage is internally consistent.

cedoor: Yes, that's something I've seen in `ethers` and some other packages, but I realized it could actually be confusing after looking at the `supportedTypes` list.

Agree it's not related to this PR, which is consistent anyway.

cedoor (approved):
LGTM, thanks @artwyman 🙏🏽

Issue #199, opened 2024-03-13
eddsa-poseidon should return numbers as bigint not string


artwyman:
**Is your feature request related to a problem? Please describe.**

All the functions in eddsa-poseidon return numeric values in stringified form, including bare `string` as well as `Point<string>` and `Signature<string>`.  In all cases, the values are `bigint` internally, and converted via `toString()` before return.

For a caller who wants to manipulate these numbers mathematically, this is a pessimization, since the caller will need to immediately parse the string back into a bigint.  

**Describe the solution you'd like**

Change these functions to return `bigint`.  Callers who want a string can call `toString()` and pay the same cost they're already paying.  Callers who want a `bigint` save that cost.

**Describe alternatives you've considered**
Leaving it as-is is the primary alternative.  It would avoid a breaking change.  The annoyance and cost is relatively small, but now is probably a better time for a breaking change than in future when there will be more users of a more stable package.

**Additional context**

I'm willing to implement this change, but this would be a breaking change, so I'm submitting this issue to spark discussion before I submit a PR.  Attn: @cedoor.

I don't know what other users of this library exist who could be broken by this.  If they're passing the values to another zk-kit function which accepts multiple input types (like a BigNumberish) their code would keep working.  If they're converting to a number via `BigInt()` that will also keep working, since the extra conversion is a nop.  However if they're using the values in a way which requires a string, they'd need to add an explicit call to toString().

cedoor:
Hey @artwyman! I think this change is totally fine and it would be more consistent with `baby-jubjub`, which is already returning bigints.

> but now is probably a better time for a breaking change than in future when there will be more users of a more stable package.

Yes, there aren't many projects using these libraries and they aren't stable yet. So this is the right time.

Feel free to open a PR 👍🏽

PR #201, opened 2024-03-14
feat(eddsa-poseidon): add packSignature and unpackSignature to eddsa-poseidon

artwyman:
## Description

Signatures are packed into 64 bytes using the same format as circomlibjs.  Unit tests confirm the compatibility of output.

## Other information

There's a TODO in place for one more unit test I'd like to add after #200 lands and fixes the ability to accept signatures in numeric form.

The unpackSignature function returns a `Signature<string>`.  If #199 gets accepted and implemented, it should change to return `Signature<bigint>`.

### DIFF
--- a/packages/eddsa-poseidon/src/eddsa-poseidon.ts
+++ b/packages/eddsa-poseidon/src/eddsa-poseidon.ts
@@ -198,8 +198,8 @@ export function packSignature(signature: Signature): Buffer {
 
     const packedR8 = packPoint(numericSignature.R8)
     const packedBytes = Buffer.alloc(64)
-    packedBytes.set(leBigIntToBuffer(packedR8), 0)
-    packedBytes.set(leBigIntToBuffer(numericSignature.S), 32)
+    packedBytes.set(leBigIntToBuffer(packedR8, 32), 0)
+    packedBytes.set(leBigIntToBuffer(numericSignature.S, 32), 32)
     return packedBytes
 }
 
--- a/packages/eddsa-poseidon/src/eddsa-poseidon.ts
+++ b/packages/eddsa-poseidon/src/eddsa-poseidon.ts
@@ -10,7 +10,7 @@ import {
     unpackPoint
 } from "@zk-kit/baby-jubjub"
 import type { BigNumberish } from "@zk-kit/utils"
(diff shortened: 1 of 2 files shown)


packages/eddsa-poseidon/src/eddsa-poseidon.ts
```diff
@@ -174,6 +174,60 @@ export function unpackPublicKey(publicKey: BigNumberish): Point<string> {
...
+    }
+
+    const numericSignature: Signature<bigint> = {
+        R8: signature.R8.map((c) => BigInt(c)) as Point<bigint>,
+        S: BigInt(signature.S)
+    }
+
+    const packedR8 = packPoint(numericSignature.R8)
+    const packedBytes = Buffer.alloc(64)
+    packedBytes.set(leBigIntToBuffer(packedR8), 0)
```
cedoor, line 201: Should `leBigIntToBuffer` also take the number of bytes as the second parameter here (i.e. size = 32)?

artwyman: Good call.  I forgot that argument was added.

artwyman: Interesting.  In attempting to write a unit test to catch a length issue here, I discovered that the code is actually correct without an explicit length.  Because it's using little-endian byte order, it's okay to have a smaller-than-32-byte buffer set into the buffer, since the rest will already be padded with zeroes at the correct end.  If we were using BE the truncation would ended up with the bytes at the wrong end.  Meanwhile numbers larger than 32 bytes already rules out by the inCurve and > subOrder checks above.

I'll add the explicit size anyway for clarity.

cedoor: Ahh good to know! Thanks for looking into it 👍🏽

cedoor (commented):
@artwyman Do you want to wait for #199 before merging this?

artwyman:
> @artwyman Do you want to wait for #199 before merging this?

No, go ahead and merge this one and I'll build my PR on top of it.

cedoor (approved):

PR #203, opened 2024-03-14
Require right types in `conversions` module

### LINKED ISSUE #187: Incorrect type assumption in bignnumberishToBigint
**Describe the bug**
There's a type-cast to Buffer at this line: https://github.com/privacy-scaling-explorations/zk-kit/blob/main/packages/utils/src/conversions.ts#L173

This is after checking for other types of bignumberish conversion.  In this case, though, it's not guaranteed the input is a buffer.  It could also be a string which doesn't match the correct format for decimal or hexadecimal encoding, meaning it's an invalid input which should throw.

I think it's going to end up throwing eventually anwyway, because a call to `n.toString("hex")` is invalid if `n` is a string.  However this should probably be checked and reported in a more explicit and clear way rather than casting to an incorrect type.

**Additional context**
Originated from discussion in #178.

cedoor:
## Description

The utility functions in the conversions module now check their parameter types before converting any value.

Parameters have also been renamed to 'value', to make them more consistent with the other modules.

Additionally, support for Uint8Array buffers has been added, when the input is a buffer.

## Related Issue(s)

Closes #187

## Checklist

### DIFF
--- a/packages/utils/src/conversions.ts
+++ b/packages/utils/src/conversions.ts
@@ -13,17 +13,19 @@
  */
 
 import { Buffer } from "buffer"
-import { requireHexadecimal, requireTypes } from "./error-handlers"
-import { isBigInt, isBuffer, isHexadecimal, isNumber, isStringifiedBigInt } from "./type-checks"
+import { requireBigInt, requireBigNumberish, requireHexadecimal, requireTypes } from "./error-handlers"
+import { isBuffer, isHexadecimal } from "./type-checks"
 import { BigNumber, BigNumberish } from "./types"
 
 /**
  * Converts a bigint to a hexadecimal string.
- * @param n The bigint value to convert.
+ * @param value The bigint value to convert.
  * @returns The hexadecimal representation of the bigint.
  */
-export function bigIntToHexadecimal(n: bigint): string {
-    let hex = n.toString(16)
+export function bigIntToHexadecimal(value: bigint): string {
+    requireBigInt(value, "value")
+
+    let hex = value.toString(16)
 
     // Ensure even length.
     if (hex.length % 2 !== 0) {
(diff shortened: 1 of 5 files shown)


packages/utils/src/conversions.ts
```diff
@@ -36,50 +38,60 @@ export function bigIntToHexadecimal(n: bigint): string {
...
 
 /**
  * Converts a buffer to a bigint. Alias for beBufferToBigInt.
- * @param b The buffer to convert.
+ * @param value The buffer to convert.
  * @returns The bigint representation of the buffer's contents.
  */
-export function bufferToBigInt(b: Buffer): bigint {
-    return beBufferToBigInt(b)
+export function bufferToBigInt(value: Buffer | Uint8Array): bigint {
```
artwyman, line 93: I think Buffer is a subtype of Uint8Array, so the use of 2 types here is redundant, though it might be a good idea anyway for clarity.

packages/utils/src/conversions.ts
```diff
@@ -148,41 +160,45 @@ export function leBigIntToBuffer(n: bigint, size?: number): Buffer {
...
  * the return value will be the bigint itself, otherwise it will be converted
  * to a bigint using big-endian byte order.
- * @param n The BigNumberish value to convert.
+ * @param value The BigNumberish value to convert.
  * @returns The bigint representation of the BigNumberish value.
  */
-export function bigNumberishToBigInt(n: BigNumberish): bigint {
-    if (isNumber(n) || isBigInt(n) || isStringifiedBigInt(n) || isHexadecimal(n)) {
-        return BigInt(n as BigNumber | number)
+export function bigNumberishToBigInt(value: BigNumberish): bigint {
```
artwyman, line 177: If you're going to accept Uint8Arrays elsewhere, it becomes a bit inconsistent that the Bignumberish type doesn't include them.

artwyman (approved):
Some comments for potential improvements, but none of them need to block this PR.

packages/utils/src/conversions.ts
cedoor: You're right! I'll add it there as well

packages/utils/src/conversions.ts
cedoor: Yes, Buffer extends Uint8Array, but I'd keep it for clarity as you said.

artwyman (approved):

PR #205, opened 2024-03-15
feat(eddsa-poseidon)!: make EdDSAPoseidon return bigints not strings

### LINKED ISSUE #199: eddsa-poseidon should return numbers as bigint not string

artwyman:
BREAKING CHANGE: Return type changes will require users to update their code.

re #199

## Description

See #199 for the argument for this change, and the discussion.

This PR changes the return types of all EdDSAPoseidon functions to bigint rather than stringified bigint.  It also updates all callers within the zk-kit repo, including in other packages.  This is a breaking change, so users of this package will probably need to update some code.  I request that maintainers update version numbers for the next release appropriately, or let me know if I should do so in this PR.

I want a sanity-check for the changes outside of eddsa-poseidon.  I was surprised (by build and test failures) to discover that other packages within the monorepo didn't depend on the latest version of eddsa-poseidon (and other packages such as utils).  I updated those dependencies, so that I wouldn't force future maintainers to update the call sites for the breaking change.  I would appreciate confirmation of whether that's the preferred best-practice in this repo.

## Related Issue(s)

Closes #199 

## Checklist


packages/eddsa-poseidon/tests/index.test.ts
```diff
@@ -224,35 +224,35 @@ describe("EdDSAPoseidon", () => {
...
     })
 
-    it("Should pack a public key (stringified)", async () => {
+    it("Should pack a public key (numeric)", async () => {
         const publicKey = derivePublicKey(privateKey)
 
         const packedPublicKey = packPublicKey(publicKey)
 
-        const expectedPackedPublicKey = babyjub.packPoint([BigInt(publicKey[0]), BigInt(publicKey[1])])
+        const expectedPackedPublicKey = babyjub.packPoint([publicKey[0], publicKey[1]])
```
cedoor, line 237:
```suggestion
        const expectedPackedPublicKey = babyjub.packPoint(publicKey)
```

packages/poseidon-cipher/tests/utils.ts
```diff
@@ -7,11 +7,7 @@ import { EncryptionKey } from "../src/types"
  * @param privateKey the private key to generate a public key from
  * @returns the public key
  */
-export const genPublicKey = (privateKey: bigint): [bigint, bigint] => {
-    const key = derivePublicKey(privateKey)
-
-    return key.map((x: string) => BigInt(x)) as [bigint, bigint]
-}
+export const genPublicKey = (privateKey: bigint): [bigint, bigint] => derivePublicKey(privateKey)
```
cedoor, line 10: Same as the other comment. `derivePublicKey` could be used directly now.

cedoor (changes requested):

cedoor:
Thanks @artwyman, left some comments and suggestions!

> I request that maintainers update version numbers for the next release appropriately, or let me know if I should do so in this PR.

I can do it after merging. I think all the packages involved are under v0, so we could increase the minor as soon as we consider these packages still unstable.

> I updated those dependencies, so that I wouldn't force future maintainers to update the call sites for the breaking change. I would appreciate confirmation of whether that's the preferred best-practice in this repo.

I wrote a comment with an assumption. I think it's ok to have them updated in this PR. From now on the inter-dependencies should be updated correctly. I'll make sure they work.

cedoor (approved):

PR #215, opened 2024-03-18
Restrict EdDSA private-key supported types

### LINKED ISSUE #188: Restrict types supported by functions in the EdDSA Poseidon package

Currently, functions such as `deriveSecretScalar`, `derivePublicKey` and `signMessage` accept different types of private keys. BigNumberish can be: `bigint`, `number`, `string`, hexadecimal (string) or Buffer.  Although supporting and converting many types reduces the dev's effort in certain contexts, it may also lead to ambiguity if a type is not correctly identified. For example, a hexadecimal string without '0x' is identified as plain text or as a bigint (if it contains only digits) depending on its value, and inconsistency remains between the length of an accepted arbitrary text (no limit at the moment) and the value of a bigint or a number, which have different size limits.

In general, the right choice for determining how many and which types to support depends on the case. Prioritizing simplicity and accepting only one type (e.g. Buffer) may be the right choice if you want to make your code more efficient and eliminate ambiguity in the interpretation of types. Making the code more flexible and adaptable may be a better choice in the case of prioritizing the user experience.

This is an implementation of an algorithm where performance matters, but at the same time the aim is to give devs an optimal experience. An intermediate solution might be to reduce the number of types to 1 or 2, and provide utilities to convert any values before passing them to library functions.

**Additional context**

Originated from discussion in https://github.com/privacy-scaling-explorations/zk-kit/pull/178#discussion_r1509782449.

cedoor:
## Description

Breaking change ⚠️ 

Types supported for the EdDSA private key have been reduced to 3: Buffer, Uint8Array, and string, i.e. buffers or text.

## Related Issue(s)

Closes #188

## Checklist

### DIFF
--- a/packages/eddsa-poseidon/src/eddsa-poseidon.ts
+++ b/packages/eddsa-poseidon/src/eddsa-poseidon.ts
@@ -11,7 +11,12 @@ import {
 } from "@zk-kit/baby-jubjub"
 import type { BigNumberish } from "@zk-kit/utils"
 import { crypto, requireBuffer } from "@zk-kit/utils"
-import { bigNumberishToBigInt, leBigIntToBuffer, leBufferToBigInt } from "@zk-kit/utils/conversions"
+import {
+    bigNumberishToBigInt,
+    leBigIntToBuffer,
+    leBufferToBigInt,
+    bufferToHexadecimal
+} from "@zk-kit/utils/conversions"
 import { requireBigNumberish } from "@zk-kit/utils/error-handlers"
 import F1Field from "@zk-kit/utils/f1-field"
 import * as scalar from "@zk-kit/utils/scalar"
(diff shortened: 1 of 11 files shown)


packages/eddsa-poseidon/src/eddsa-poseidon.ts
```diff
@@ -26,12 +31,15 @@ import { hash as blake, checkMessage, checkPrivateKey, isPoint, isSignature, pru
  * and converting it into a little-endian integer. The use of the secret scalar streamlines the public key generation
  * process by omitting steps 1, 2, and 3 as outlined in RFC 8032 section 5.1.5, enhancing circuit efficiency and simplicity.
  * This method is crucial for fixed-base scalar multiplication operations within the correspondent cryptographic circuit.
+ * The private key must be an instance of Buffer, Uint8Array or a string. If you want to pass a bigint, a number or a
```
artwyman, line 34: Be explicit about how a string is interpreted in this case.  I.e. the string is used as a set of raw bytes (in UTF-8, I believe) as input to key generation.  It's not interpreted as hex, or decimal, or anything else.

Also probably be explicit in saying the input can be any size, and all bytes will be used to generate entropy (vs. other inputs where a Buffer is expected to represent a number of a specific size).

artwyman: Same comment in all the other functions which take private key as input, so I won't duplicate this comment.

artwyman (approved):
Looks good functionally, but see my comment for a suggested clarification to docs.

0xjei (approved):

packages/eddsa-poseidon/src/eddsa-poseidon.ts
cedoor: Thanks! I updated them 👍🏽

PR #219, opened 2024-03-20
Use `if` instead of `while` to increment depth in the `LeanIMT.insert` function

### LINKED ISSUE #213: Use `if` instead of `while` to increment `depth` in `InternalLeanIMT._insert()`
A new insertion can increase tree's depth by at most 1. We can be sure that this loop isn't executed twice:
https://github.com/privacy-scaling-explorations/zk-kit/blob/45739f6bed1acf31d8791b728346ea6ae9092915/packages/imt.sol/contracts/internal/InternalLeanIMT.sol#L49-L51

Thus, the `while` loop can be turned into an `if` condition saving gas. Add a comment explanation too.

**Additional context**
PSE audit.

cedoor:
## Description

A new insertion can increase a tree's depth by at most 1. We can be sure that the while loop isn't executed twice.

## Related Issue(s)

Closes #213

## Checklist

### DIFF
--- a/packages/imt.sol/contracts/internal/InternalLeanIMT.sol
+++ b/packages/imt.sol/contracts/internal/InternalLeanIMT.sol
@@ -46,6 +46,9 @@ library InternalLeanIMT {
             revert LeafAlreadyExists();
         }
 
+        // A new insertion can increase a tree's depth by at most 1,
+        // and only if the number of leaves supported by the current
+        // depth is less than the number of leaves to be supported after insertion.
         if (2 ** self.depth < self.size + 1) {
             self.depth += 1;
         }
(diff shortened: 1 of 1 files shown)


packages/imt.sol/contracts/internal/InternalLeanIMT.sol
```diff
@@ -46,7 +46,7 @@ library InternalLeanIMT {
             revert LeafAlreadyExists();
         }
 
-        while (2 ** self.depth < self.size + 1) {
```
0xbok, line 49: maybe add a comment here to explain why checking just once is enough

cedoor: Sure, makes sense!

cedoor: @0xbok could you check if it's ok?

0xbok (approved):

vplasencia (approved):

PR #223, opened 2024-03-21
feat(imt): add static version of IMT verifyProof

### LINKED ISSUE #218: IMT's verifyProof function should be static

The verifyProof function in both IMT (https://github.com/privacy-scaling-explorations/zk-kit/blob/main/packages/imt/src/imt/imt.ts#L299) and LeanIMT (https://github.com/privacy-scaling-explorations/zk-kit/blob/main/packages/imt/src/lean-imt/lean-imt.ts#L275) don't actually reference any state in the tree other than the `_hash` function.  I suggest separating this function from needing to be called on an instance object for two reasons:

- It's somewhat inconvenient to have to create a tree object in order to validate an unrelated proof.
- It's somewhat misleading that validating the proof checks the root in the proof, but not the root in the tree.  Without reading the code, it would be easy to assume it's validating that an element exists in this particular tree.

I'd suggest making a static function (or non-class function) which takes the hash as an argument.  There can also be a convenience function on the class itself which uses `this._hash` to call the static function.

Modifying the existing function to be static would be a breaking change, but if the static/free function is given a different name there's no need for that.

I can submit a PR for this after any discussion or feedback on the idea.

**Additional context**
This came up in the Zupass repo, where I ended up working around it by creating a new empty tree here: https://github.com/proofcarryingdata/zupass/pull/1564/files#diff-611d148c3a78ab428ec4a7fbb523eefddcf720a9833b214a25253b5b7ec651efR224

artwyman:
re #218

## Description

Adds static versions of `verifyProof` on `IMT` and `LeanIMT` so that it's not necessary to create an IMT object just to verify an existing proof.  Existing instance method still exists (to avoid breaking change) and calls the static one.  Also clarified in docs that the instance method doesn't check that the proof comes from this specific tree (i.e. no checking of root).

## Related Issue(s)

Closes #218

## Checklist

### DIFF
--- a/packages/imt/src/imt/imt.ts
+++ b/packages/imt/src/imt/imt.ts
@@ -292,11 +292,24 @@ export default class IMT {
 
     /**
      * It verifies a {@link IMTMerkleProof} to confirm that a leaf indeed
-     * belongs to the tree.
+     * belongs to a tree.  Does not verify that the node belongs to this
+     * tree in particular.  Equivalent to `IMT.verifyProof(proof, this._hash)`.
+     *
      * @param proof The Merkle tree proof.
      * @returns True if the leaf is part of the tree, and false otherwise.
      */
     public verifyProof(proof: IMTMerkleProof): boolean {
+        return IMT.verifyProof(proof, this._hash)
+    }
+
+    /**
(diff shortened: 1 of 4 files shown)


cedoor (approved):
Looks great to me! Thanks @artwyman

PR #240, opened 2024-04-08
refactor: move `download` and `getSnarkArtifacts` into `@zk-kit/utils`

### LINKED ISSUE #176: Move `eddsa-proof` and `poseidon-proof` packages `get-snark-artifacts.node.ts` duplicated methods to `@zk-kit/utils`
Consider generalising the `download()` and `getSnarkArtifacts()` functions in the `eddsa-proof` and `poseidon-proof` packages and exporting them in a generic utility package such as `@zk-kit/utils`. This will avoid repetition and make the functions more easily reusable.

**Additional context**
Please refer to the following link for more information: https://github.com/privacy-scaling-explorations/zk-kit/pull/173#issuecomment-1956509325

sripwoud:
## Description
To avoid repetition, this PRs moves the function related to downloading and getting the file path of the snark artifacts out of `@zk-kit/{eddsa,poseidon}-proof` and into `@zk-kit/utils` instead.  

Additionally it:
- does the same for [semaphore](https://github.com/semaphore-protocol/semaphore/blob/main/packages/proof/src/get-snark-artifacts.node.ts) (once new `@zk-kit/utils` is published, I will submit changes in the semaphore repo too)
- defines conditional types/function overloading in order to be able to encapsulate logic in a factory function (in case other proofs needs to be added in the future, it should be straightforward) while still having nicer type checking
   ```
   maybeGetPoseidonSnarkArtifacts(2) // ok
   maybeGetPoseidonSnarkArtifacts() // ts error
   maybeGetEdDSASnarkArtifacts() // ok
   maybeGetEdDSASnarkArtifacts(2) // ts error
   maybeGetSemaphoreSnarkArtifacts(2) // ok
   maybeGetSemaphoreSnarkArtifacts() // ts error
   ```
- improves error handling
- fetches artifacts from unpkg (see https://github.com/privacy-scaling-explorations/snark-artifacts)
- writes unit tests for these functions
  image

**It introduces breaking changes regarding the packages exporting the functions and the function names.**

|Before|After|
|---|---|
|`import { getSnarkArtifacts } from @zk-kit/eddsa-proof`<br>`import { getSnarkArtifacts } from @zk-kit/poseidon-proof`|`import { maybeGetEddsaSnarkArtifacts, maybeGetPoseidonSnarkArtifacts } from @zk-kit/utils`|

## Related Issue(s)
Closes #176 

## Checklist


package.json
```diff
@@ -65,7 +65,7 @@
         "eslint-plugin-import": "^2.29.1",
         "eslint-plugin-jest": "^27.9.0",
         "husky": "^9.0.11",
-        "jest": "^29.7.0",
+        "jest": "30.0.0-alpha.1",
```
sripwoud, line 68: in order to be able to use `matchInlineSnapshot` matchers
https://github.com/jestjs/jest/pull/14566

packages/utils/src/get-snark-artifacts/get-snark-artifacts.node.ts
```diff
@@ -0,0 +1,98 @@
...
+
+            fileStream.write(Buffer.from(value))
+            await pump()
+        }
+
+        await pump()
+    } catch (error) {
+        fileStream.close()
+        throw error
+    }
```
sripwoud, line 60: even if node > 18 has built in fetch. There are still some mismatch between the stream api of node and [whatwg](https://streams.spec.whatwg.org/): node fetch api returns a [ReadableStream](https://streams.spec.whatwg.org/#rs-class) which isn't directly compatible with node stream system. This is why `as any` type casting was necessary.
We can avoid type casting by using the stream reader from the ReadableStream which is part of the WHATWG api.  
It also makes unit testing easier.  
But it is lot more verbose....

packages/utils/tests/get-snark-artifacts.node.test.ts
```diff
@@ -0,0 +1,127 @@
...
+})
+
+describe("getPoseidonSnarkArtifacts", () => {
+    it("should handle fetch errors", async () => {
+        ;(existsSync as jest.Mock).mockReturnValue(false)
+        ;(global.fetch as jest.Mock).mockResolvedValueOnce({
+            ok: false,
+            statusText: "TEST"
+        })
+        await expect(getPoseidonSnarkArtifacts(2)).rejects.toThrowErrorMatchingInlineSnapshot(
```
sripwoud, line 30: needs jest >= 30.  
see https://github.com/privacy-scaling-explorations/zk-kit/pull/240/files#r1556087134

somehow I could not make it work with [`prettierPath`](https://jestjs.io/docs/configuration#prettierpath-string)

sripwoud:
@cedoor I may need a little help to wrap that PR. I am not super familiar with rollup. I am not sure how to update the rollup config files following my changes.
Especially I think I don't configure rollup to build correctly to `eddsa-proof` (only the tests for eddsa-proof fail, I believe it is because of the broken build).
`yarn build` for `eddsa-proof` only outputs to stdout ...

packages/utils/src/get-snark-artifacts/get-snark-artifacts.node.ts
```diff
@@ -0,0 +1,89 @@
...
```
sripwoud, line 52: even if node > 18 has built in fetch, there are still some mismatch between the stream api of node and [whatwg](https://streams.spec.whatwg.org/): node fetch api returns a [ReadableStream](https://streams.spec.whatwg.org/#rs-class) which isn't directly compatible with node stream system. This is why `as any` type casting was necessary.
We can avoid type casting by using the stream reader from the ReadableStream which is part of the WHATWG api.  
It also makes unit testing easier.  
But it is lot more verbose....

package.json
cedoor: How unstable is it? `alpha.1` sounds scary :D

packages/utils/src/get-snark-artifacts/get-snark-artifacts.node.ts
cedoor: Sounds good to me! How much of a decrease in performance?

packages/eddsa-proof/src/generate.ts
```diff
@@ -30,7 +29,7 @@ export default async function generate(
     // If the Snark artifacts are not defined they will be automatically downloaded.
     /* istanbul ignore next */
     if (!snarkArtifacts) {
-        snarkArtifacts = await getSnarkArtifacts()
+        snarkArtifacts = await getEddsaSnarkArtifacts()
```
cedoor, line 32: Maybe better `EdDSA`? just to make it consistent with the same occurrence in the other packages.

packages/utils/src/get-snark-artifacts/config.ts
```diff
@@ -0,0 +1,8 @@
+import { ArtifactType, ProofType } from "../types"
+
+export const ARTIFACTS_BASE_URL = "https://zkkit.cedoor.dev"
+export const ARTIFACTS_TYPES = [ArtifactType.WASM, ArtifactType.ZKEY]
+export const ARTIFACT_BASE_URLS: Record<ProofType, string> = {
+    [ProofType.POSEIDON]: `${ARTIFACTS_BASE_URL}/${ProofType.POSEIDON}-proof/artifacts`,
```
cedoor, line 6: Semaphore is also using the same functions: https://github.com/semaphore-protocol/semaphore/blob/main/packages/proof/src/get-snark-artifacts.node.ts

We could add support for it here as well, maybe using the same ZK-Kit base URL (?)

package.json
sripwoud: none of the existing tests broke, so I it looks ok I'd say.
Looking at the [breaking changes](https://github.com/jestjs/jest/releases/tag/v30.0.0-alpha.1), it does not look like impacting how zk-kit uses jest.
There is `alpha.3` released since.
As it is "only" a devDep, I'd YOLO merge it ^^.
Your call. I can also stop using the snapshot matchers in the tests I added too...

cedoor: I'm in, you convinced me ahah 👍🏽

packages/utils/src/get-snark-artifacts/get-snark-artifacts.node.ts
sripwoud: Actually, and unexpectedly, an increase of performance (probably due to the fact that I am fetching with parallel with `Promise.all` too?)

benchmark on poseidon. Run it 2 times. Gave similar results both times
```
                   pipe pump
poseidon 1 inputs: 2389 44
poseidon 2 inputs: 690 1
poseidon 3 inputs: 723 1
poseidon 4 inputs: 672 1
poseidon 5 inputs: 707 3
poseidon 6 inputs: 915 1
poseidon 7 inputs: 1653 1
poseidon 8 inputs: 722 1
poseidon 9 inputs: 878 1
poseidon 10 inputs: 772 1
poseidon 11 inputs: 790 1
poseidon 12 inputs: 806 1
poseidon 13 inputs: 819 1
poseidon 14 inputs: 851 1
poseidon 15 inputs: 776 1
poseidon 16 inputs: 871 1
```

packages/utils/src/get-snark-artifacts/config.ts
sripwoud: good idea. Will do it

sripwoud:
@cedoor I need some help to wrap that PR.
I am struggling to update the rollup config (not super familiar with rollup), especially splitting the config files between node/browser. I modified the config of the utils, eddsa-proof, poseidon-proof packages.
The utils build and tests run fine.  
But I broke the build of eddsa-proof and poseidon-proof (even locally, it just outputs to STDOUT).
I am not sure what I am doing wrong.

packages/utils/src/get-snark-artifacts/config.ts
```diff
@@ -0,0 +1,75 @@
+import { Artifact, Proof } from "../types"
+
+export const URLS = {
+    zkkit: "https://zkkit.cedoor.dev",
+    semaphore: "https://semaphore.cedoor.dev"
+}
+
+export const ARTIFACTS = Object.values(Artifact)
+
+export function GetSnarkArtifactUrl({
```
sripwoud, line 10: start of ts function overloads

packages/utils/src/get-snark-artifacts/config.ts
sripwoud, line 1: Btw, this whole thing is a bit complicated because files are hosted on different urls.
Maybe it makes sense to have everything on eg https://artifacts.cedoor.dev/

Then https://artifacts.cedoor.dev/{eddsa,poseidon,semaphore}/...

?

packages/utils/src/get-snark-artifacts/get-snark-artifacts.node.ts
cedoor: Ohh great news 👍🏽

packages/utils/src/get-snark-artifacts/config.ts
cedoor: Yes, I think it makes sense to have all of them in the same URL

cedoor: I'll upload the Semaphore artifacts on `https://zkkit.cedoor.dev` asap. I'll let you know here.

cedoor:
> @cedoor I think I need some little help to wrap that PR. I am struggling to update the rollup config (not super familiar with rollup), especially splitting the config files between node/browser. I modified the config of the utils, eddsa-proof, poseidon-proof packages. The utils build and tests run fine. But I broke the build of eddsa-proof and poseidon-proof (even locally, it just outputs to STDOUT). I am not sure what I am doing wrong.

I think you just need to update the `package.json` file. The `exports` field doesn't need browser/node paths anymore. It should be like the `baby-jubjub/package.json` file now.

sripwoud:
the workflow log shows that the failing tests aren't related to this PR https://github.com/privacy-scaling-explorations/zk-kit/actions/runs/8723164679/job/23930749835?pr=240#step:8:1069 (LazyIMT)

packages/eddsa-proof/tests/index.test.ts
```diff
@@ -14,6 +14,7 @@ describe("EddsaProof", () => {
 
     beforeAll(async () => {
         curve = await buildBn128()
+        fullProof = await generate(privateKey, scope)
```
sripwoud, line 17: if accessing an outer scope variable in both tests, we need to make sure it is setup before both of them. As jest runs tests concurrently, I have experienced failing tests because the 2nd test would run before the 1st and fullProof wasn't defined yet.

packages/poseidon-proof/tests/index.test.ts
```diff
@@ -43,48 +43,46 @@ const poseidonFunctions = [
...
-            }
-        }, 60000)
+    it("Should verify a Poseidon proof from 1 to 16 preimage(s)", async () => {
+        proofs.forEach(async ({ fullProof }, i) => {
+            await expect(verify(i + 1, fullProof)).resolves.toBe(true)
+        })
+    })
 
-        it("Should verify an invalid Poseidon proof", async () => {
-            fullProof.digest = "3"
```
sripwoud, line 83: like the eddsa test. Here we access `fullProof` defined in the outer scope but set by the first test. It is not a reliable way to set `fullProof` as we can't assume jest will always finish the first test before the second one. Better setting everything cleanly in a `beforeAll` hook

packages/utils/rollup.config.ts
```diff
@@ -24,7 +24,7 @@ export default [
             { file: pkg.exports["."].node.require, format: "cjs", banner, exports: "auto" },
             { file: pkg.exports["."].node.default, format: "es", banner }
         ],
-        external: [...Object.keys(pkg.dependencies), "crypto"],
+        external: [...Object.keys(pkg.dependencies), "crypto", "node:fs", "node:fs/promises", "node:path", "node:os"],
```
sripwoud, line 27: this suppresses some rollup warning. Not 100% it is correct to fix it like this?

cedoor: Yes, that's what I've done with the other packages

packages/eddsa-proof/tests/index.test.ts
cedoor: Umh I guess it would be better to have independent tests in general. Generating a proof takes some time tho, so in this case it could make sense. Isn't it possible to run those tests sync?

packages/eddsa-proof/package.json
```diff
@@ -54,7 +48,7 @@
         "@ethersproject/bytes": "^5.7.0",
         "@ethersproject/keccak256": "^5.7.0",
         "@zk-kit/eddsa-poseidon": "0.11.0",
-        "@zk-kit/utils": "1.0.0-beta.1",
+        "@zk-kit/utils": "workspace:^",
```
cedoor, line 51: This won't work when publishing the package. It should be `1.0.0-beta.2`

packages/poseidon-proof/package.json
```diff
@@ -53,6 +47,7 @@
         "@ethersproject/bignumber": "^5.7.0",
         "@ethersproject/bytes": "^5.7.0",
         "@ethersproject/keccak256": "^5.7.0",
+        "@zk-kit/utils": "workspace:^",
```
cedoor, line 50: Same here, it needs the explicit version

packages/poseidon-proof/tests/index.test.ts
```diff
@@ -43,48 +43,46 @@ const poseidonFunctions = [
...
+let curve: any
+const proofs: Array<{ fullProof: PoseidonProof; digest: bigint }> = []
 
-    let fullProof: PoseidonProof
-    let curve: any
+beforeAll(async () => {
+    curve = await buildBn128()
+    for (const preimage of preimages) {
+        const currentPreimages = preimages.slice(0, preimage)
+        const fullProof = await generate(currentPreimages, scope)
```
cedoor, line 55: Do we really need to test all the instances here?

packages/poseidon-proof/src/generate.ts
```diff
@@ -26,19 +25,18 @@ export default async function generate(
     scope: BytesLike | Hexable | number | bigint,
     snarkArtifacts?: SnarkArtifacts
 ): Promise<PoseidonProof> {
-    // If the Snark artifacts are not defined they will be automatically downloaded.
-    /* istanbul ignore next */
-    if (!snarkArtifacts) {
-        snarkArtifacts = await getSnarkArtifacts(preimages.length)
-    }
+    // allow user to override our artifacts
+    // if not explicitly provided: download them from unpkg if not already in local tmp folder
```
cedoor, line 29: Maybe worth not mentioning unpkg here as it could change. External code shouldn't know what kind of CDN is used under the hood. Also, it'd be hard to update all the external occurences.

packages/utils/src/get-snark-artifacts/config.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    switch (proof) {
+        case Proof.EDDSA:
+            return {
+                [Artifact.WASM]: `${BASE_URL}/${proof}.${Artifact.WASM}`,
+                [Artifact.ZKEY]: `${BASE_URL}/${proof}.${Artifact.ZKEY}`
+            }
+
+        case Proof.POSEIDON:
+            if (numberOfInputs === undefined) throw new Error("numberOfInputs is required for Poseidon proof")
```
cedoor, line 67: `@zk-kit/utils` provides some utility functions to check if parameters are defined or have a specific type. We could consider using them here.

packages/utils/src/get-snark-artifacts/config.ts
```diff
@@ -0,0 +1,87 @@
...
+export async function GetSnarkArtifactUrls({
+    proof,
+    numberOfInputs,
+    version
+}: {
+    proof: Proof.POSEIDON
+    numberOfInputs: number
+    version?: Version
+}): Promise<SnarkArtifacts>
+export async function GetSnarkArtifactUrls({
```
cedoor, line 26: Why are these functions capitalized?

packages/utils/tests/config.test.ts
cedoor, line 1: In order to be consistent with the other files here, this file name should be `get-snark-artifacts.test.ts`.

I also wonder what you think about naming it `snark-artifacts`. The other modules don't use verbs, all of them are nouns.

packages/utils/tests/get-snark-artifacts.node.test.ts
cedoor, line 1: Could we have all the tests of a module in 1 test file? Just to be consistent with the other test files.

cedoor (changes requested):

packages/eddsa-proof/package.json
sripwoud: I was wondering how you handle these cases?
If the ci tests depends on a local package that hasn't been published yet... do you release it manually under beta yourself first?

packages/utils/src/get-snark-artifacts/config.ts
sripwoud: To indicate factory functions (comparable to a class constructor).
I can lowercase it

packages/poseidon-proof/tests/index.test.ts
sripwoud: Well each involves different artifacts. So I'd say yes?

packages/eddsa-proof/package.json
cedoor: Even if you specify its version, it always uses the local version locally. So the CI tests should use local bundles generated with `yarn build:libraries`.

packages/poseidon-proof/tests/index.test.ts
cedoor: Right, just wondering how much time it needs.

packages/utils/src/get-snark-artifacts/config.ts
cedoor: Ahhh, as you prefer!

packages/poseidon-proof/tests/index.test.ts
sripwoud: timeout was already set at 1min. So I kept that value

packages/utils/src/get-snark-artifacts/config.ts
sripwoud: https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBAMwK4DsDGMCWEV0wZwBFgFMVgATACgDcBDAGyWAC446UBPASjYCMIEBsA5wA3gCg40uFGAwkUXDE5hgEBHHpNgcAIQBeA3ABEqCiTKUTEgL4SJoSLESoM2XHICOSTHOKk5NRgdFB0ALbywFAAaozMbBycADRwIWGRMNEAchGscPgwUGQA5tziUjKYmlR6BAFWwaF5WbHxwNzlkjI9cDAAFlAQAO5w5KMAKqrAAKJQQ1BUAAYACs2Z0XAA5AAkYuktOXm2W3j4YxDwFoGUS9yV0vb2EsjoWDiIglR8oQD8LCgkOE+NEug9ZMAfH5gA0gt9QqkTD8oCZ7j1qnB4VA4AAeOAARnKAyGo3GcDmCyoJnCSEKcBBcAAfATUXYJEA

Unfortunately although `requireDefined` ensures the value is defined, it appears that in this case TS isn't smart enough and can't statically analyze the code flow to understand that `numberOfInputs` can't be defined, and it errors with `'numberOfInputs' is possibly 'undefined'`. So I'll keep the explicit check.

packages/utils/tests/config.test.ts
sripwoud: sure let's keep following this convention

cedoor (changes requested):
@sripwoud Great work, thank you so much for this PR 🙏🏽  I think the code is ready!

May I just ask you to add some JSDoc documentation to the functions you've added to the `@zk-kit/utils` package?

The tests that are failing are not related to this PR, so it can be merged.

sripwoud:
> @sripwoud Great work, thank you so much for this PR 🙏🏽 I think the code is ready!
> 
> May I just ask you to add some JSDoc documentation to the functions you've added to the `@zk-kit/utils` package?
> 
> The tests that are failing are not related to this PR, so it can be merged.

I have updated the documentation. Ex:
image

I did another small change: not including `GetSnarkArtifactUrls` from `config.ts` in the utils package exports. I don't think we should expose it. This is only a lower level factory function used by the other function we export (`mayGet*SnarkArtifacts`).

And I opened #258

cedoor (approved):

cedoor:
@sripwoud Great!! I think you just need to merge changes from the main branch and update the utils dependency. A new version has been released in the meantime.

sripwoud:
These tests really take an unforeseeable amount of time. I keep having to increase the timeouts. :sweat_smile:

cedoor:
> These tests really take an unforeseeable amount of time. I keep having to increase the timeouts. 😅

Ummh, I wonder whether it makes sense to test all configurations. If one works the lib should be fine. The other artifacts should be ok if the ceremony has correctly been run and certified.

PR #234, opened 2024-03-28
feat(circuits): add tests and Noir circuits for pedersen and poseidon2 proving

signorecello:
## Description

Added two more circuits to Noir's SMT proving circuits, namely with Poseidon2 and Pedersen. The only one there was using a very unoptimized poseidon1 hash. This makes SMT proving several times faster (3-5s).

Of course, to go with it, I also added tests to `@zk-kit/smt`. I refactored them with the handy Jest `.each` so it tests the three hashes at once 🙂 

## Other information

Hopefully a nice beginning of Aztec >< PSE collaboration in adding more Noir stuff to this repository!

## Checklist

-   [x] My code follows the style guidelines of this project
-   [x] I have performed a self-review of my code
-   [x] I have commented my code, particularly in hard-to-understand areas
-   [x] I have made corresponding changes to the documentation
-   [x] My changes generate no new warnings
-   [x] I have run `yarn prettier` and `yarn lint` without getting any errors
-   [x] I have added tests that prove my fix is effective or that my feature works
-   [x] New and existing unit tests pass locally with my changes
-   [x] Any dependent changes have been merged and published in downstream modules

PR #244, opened 2024-04-09
feat(circuits): add a noir circuit for merkle tree membership and JS tests for poseidon2 and pedersen

signorecello:
Follow-up to #234, this PR:

- Adds a merkle tree membership circuit in Noir. 
- Adds an abstraction layer to both MTs to make it easier to simply call for example `mt.add` to prove addition. Sweet Noir traits!
- Adds a bunch of noir tests
- Refactors the LeanIMT tests just like I did for #234, featuring `poseidon`, `poseidon2` and `pedersen` hashes.
- Abstracts tree sizes away by making the libs use slices instead. This makes them useable with any array size.
- Merges all the Merkle Trees implementations and tests into one single lib. This makes most of the code reusable, and makes the interface look similar.
- Adds READMEs with instructions


vplasencia (approved):
Great work. Thank you very much.

signorecello:
Seems like I cannot merge, can you do it please @vplasencia ?

vplasencia:
Hey @signorecello! At least two people should review the PR before merging. When another of the reviewers reviews it, we will merge it. Thanks.

cedoor (approved):

Issue #238, opened 2024-04-08
Use noble cryptography libraries


0xbok:
Noble is a set of libraries providing "Audited & minimal JS implementation" of several cryptographic primitives implemented in zk-kit.

Some links: https://github.com/paulmillr/noble-curves, https://github.com/paulmillr/noble-hashes, https://paulmillr.com/noble/

It can be used to define babyjubjub curve, import poseidon, blake and sha hashes.

As discussed with @cedoor already, it's worthwhile to use these libraries as they are audited and optimized.

cedoor:
@artwyman, @ctrlc03 tagging you here as this would be a big change that will affect the following packages:

* `@zk-kit/baby-jubjub`: It could use the [`edwards`](https://github.com/paulmillr/noble-curves/blob/main/src/abstract/edwards.ts) module, similarly to [`jubjub`](https://github.com/paulmillr/noble-curves/blob/main/src/jubjub.ts), but with its parameters;
* `@zk-kit/eddsa-poseidon`: It would use a new `@zk-kit/poseidon` package, based on https://github.com/paulmillr/noble-curves/blob/main/src/abstract/poseidon.ts.

There shouldn't be breaking changes here tho.

Curious to know what you think about this.

artwyman:
For non-breaking changes, my primary concerns to understand would be the impact on performance and bundle download size.  circomlibjs has a very large set of poseidon constants for all sizes, so switching to poseidon-lite was a big win for bundle size.  I'd be curious if these noble libraries have been optimized for that.

For breaking changes, I'm most concerned about cryptographic compatibility, rather than simple API changes.  We have lots of EdDSA tickets already in the wild which need to remain verifiable.  Those were created using circomlibjs, but I've tested compatibility and intend to switch to zk-kit for them.  There aren't yet any PODs (new PCD format for tickets and other things) in the wild, but there will be soon (early May) so changes to the cryptographic algorithms would need to either happen quickly, or we'd need a backward-compatible option there as well.  These two concerns together might argue for providing both hash options (blake1 vs. sha-3) as different configurations in the library.  It would be okay if we had to make "breaking" API changes to opt for the blake1 hash in use cases which require it, but it would be important to be able to support both hashes in the same codebase.

[...]

PR #245, opened 2024-04-09
New Poseidon library

cedoor:
## Description

> [!NOTE]
> No need for in-depth review here, but still required to merge. Open to any discussion of course.

This PR originated from #238, a proposal for using Noble, an audited library with many implementations that can be used on ZK-Kit to make code more secure.

This PR aimed to verify whether the performance and size of the bundle were equal to or better than the alternatives. The work led to the following conclusion: Noble seems solid, but the abstract [implementation of Poseidon](https://github.com/paulmillr/noble-curves/blob/main/src/abstract/poseidon.ts), which allows implementations with arbitrary parameters and constants to be built on top of it, still seems slow in comparison to [`circomlibjs`](https://github.com/iden3/circomlibjs) and [`poseidon-lite`](https://github.com/vimwitch/poseidon-lite). The bundle size is the same as that of `poseidon-lite`.

However, this branch could still be merged and remain in an experimental state (the package will not be shown in the readme nor published), pending any progress.

A future work that might be worth spending time on is to provide a version of Poseidon with the optimization introduced by Filecoin in [Neptune](https://github.com/lurk-lab/neptune), which could increase performance by sacrificing the bundle size. 

## Results

| name                                    | ops | margin | samples | promise              | min                  | max                  | mean                 | median                |   marginOfError       |
|-----------------------------------------|-----|--------|---------|----------------------|----------------------|----------------------|----------------------|-----------------------|---------------------|
| PoseidonLite v0.2.0 - 50 hashes         | 128 | 1.5    | 82      | 0.007449643571428571 | 0.012513912857142857 | 0.007838777611498256 | 0.007758738071428572 | 0.0005438473932657464 |   1.5016815772915084  |  
| CircomlibJS Poseidon v0.0.8 - 50 hashes | 116 | 1.56   | 77      | 0.008292171428571428 | 0.0132828536         | 0.00858911488262214  | 0.008471577833333334 | 0.0006014991010496103 |   1.5642180021499086  |   
| ZK-Kit Poseidon - 50 hashes             | 78  | 0.33   | 68      | 0.012469885          | 0.013173915          | 0.012799256480147065 | 0.012803638499999999 | 0.0001790579589654917 |   0.33251443458087543 |   

## Related Issue(s)

Re #238

## Checklist

### DIFF
--- a/packages/poseidon/package.json
+++ b/packages/poseidon/package.json
@@ -126,6 +126,6 @@
     },
     "dependencies": {
         "@noble/curves": "1.4.0",
-        "@zk-kit/utils": "1.0.0-beta"
+        "@zk-kit/utils": "1.2.0"
     }
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -3308,7 +3308,7 @@ __metadata:
     "@rollup/plugin-node-resolve": "npm:^15.2.3"
     "@rollup/plugin-terser": "npm:^0.4.4"
     "@rollup/plugin-typescript": "npm:^11.1.6"
-    "@zk-kit/utils": "npm:1.0.0-beta"
+    "@zk-kit/utils": "npm:1.2.0"
     circomlibjs: "npm:0.0.8"
     rimraf: "npm:^5.0.5"
     rollup: "npm:^4.12.0"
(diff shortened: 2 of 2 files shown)


artwyman:
Check if I'm reading your performance chart right: It seems that the mean is much slower than the other libaries, but the median is much faster.  If that's accurate it could be hiding a first-time setup cost issue, which might be reasonable to accept or address separately.

However it looks like the min/max values don't make sense (maybe they're reversed) and the mean/median is larger than the max in particular in most cases, so I don't know if I trust this chart in general.  Can you check that it's being generated correctly?

The structure of the library seems reasonable (similar to poseidon-lite), but I haven't checked over all the code in detail.

paulmillr:
poseidon in noble-curves now implements sponge - I believe very few libraries do this. Most only have hades permutation.

babyjubjub is available here: https://github.com/paulmillr/noble-curves/blob/bee1ffe0000095f95b982a969d06baaa3dd8ce73/src/misc.ts#L42-L52. There is no need to reimplement it

PR #247, opened 2024-04-13
refactor(utils): update F1Field.e(res) function to remove the comparison checks with this._order

aakansha1234:
F1Field.e(res) checks if it is >= this._order to determine if it needs to take a modulus with this._order. However, in v8 bigint implementation, this check is already present. So this check is not necessary as bigint implementation already does that. This commit removes these checks.

https://github.com/v8/v8/blob/497d8573dc80b1b69052a834bec894cf5d4238e7/src/builtins/builtins-bigint.tq#L380-L383

I was just reading zk-kit's F1Field code after studying modular arithmetic. Please let me know if this is useful!

## Checklist

### DIFF
--- a/packages/utils/src/f1-field.ts
+++ b/packages/utils/src/f1-field.ts
@@ -35,17 +35,8 @@ export default class F1Field {
      * @returns The equivalent value within the field.
      */
     e(res: bigint): bigint {
-        if (res < 0) {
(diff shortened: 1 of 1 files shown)


cedoor (approved):

cedoor:
@aakansha1234 Thank you vm!

PR #249, opened 2024-04-17
fix(utils): fix pow for negative exponent

0xbok:
`pow` fn comment says it handles negative exponent but it doesn't. This PR fixes it. Here's the reasoning behind the fix:

$$
a^{-e} = (a^{-1})^e \mod p
$$

So `pow(a, -e) == pow(inv(a), e)`


0xbok:
note that the pow test uses 12 as the order. A field with a non-prime order may not have an inverse for every element in the set.

cedoor (approved):

PR #257, opened 2024-04-23
Add mod reduction in secret scalar generation

### LINKED ISSUE #239: Private key can be out of range
so `l` is of 251 bits (as used in [eip-2494](https://eips.ethereum.org/EIPS/eip-2494#base-point)) (`r` in geometry's report; to be linked). The right shift by 3 bits make the private key outside this range.

https://github.com/privacy-scaling-explorations/zk-kit/blob/ce826a9f754d8efa39993647f066dbec1d2b5093/packages/eddsa-poseidon/src/eddsa-poseidon.ts#L56

Proposed fixe (after internal discussion with @gubsheep and semaphore team):
- Take $\mod l$ on top of existing implementation. This ensures that the same public key is generated for the same input to the fn.

ignore below (keeping here for history):
Previous proposed fix:
- So the private key derived has to be right shifted by at least 5 bits to get to 251 bits. Even then there's a very low non-zero chance that the key may be  >= `l`.

- (2^251-1) - r is 7.824... . So shift the key by 6 bits if you want to avoid this too. Else, you can shift it by 5 bits, and subtract 8 to get it in range.

cedoor:
## Description

The secret scalar is used to derive the public key and it must be > l, where l is the sub-order used in Baby Jubjub. Adding a mod reduction ensures it always stays in the right range by keeping it compatible with the existing public keys.

## Related Issue(s)

Fixes #239

## Checklist

### DIFF
--- a/packages/eddsa-poseidon/src/eddsa-poseidon.ts
+++ b/packages/eddsa-poseidon/src/eddsa-poseidon.ts
@@ -53,7 +53,7 @@ export function deriveSecretScalar(privateKey: Buffer | Uint8Array | string): bi
     hash = hash.slice(0, 32)
     hash = pruneBuffer(hash)
 
-    return scalar.shiftRight(leBufferToBigInt(hash), BigInt(3))
+    return scalar.shiftRight(leBufferToBigInt(hash), BigInt(3)) % subOrder
 }
 
 /**


packages/eddsa-poseidon/src/eddsa-poseidon.ts
```diff
@@ -53,7 +53,7 @@ export function deriveSecretScalar(privateKey: Buffer | Uint8Array | string): bi
     hash = hash.slice(0, 32)
     hash = pruneBuffer(hash)
 
-    return scalar.shiftRight(leBufferToBigInt(hash), BigInt(3))
+    return scalar.shiftRight(leBufferToBigInt(hash), BigInt(3)) % subOrder
```
0xbok, line 56: can you confirm `subOrder` is `2736030358979909402780800718157159386076813972158567259200215660948447373041`?

I think so but still good to confirm from you.

cedoor: Just double-checked! It's right

PR #270, opened 2024-04-25
refactor!: change types for scope parameter

### LINKED ISSUE #214: `scope` parameter in `@zk-kit/eddsa-proof` and `@zk-kit/poseidon-proof` could also support text

cedoor:
## Description

This PR changes the type of the `scope` parameter in the following packages: `@zk-kit/eddsa-proof`, `@zk-kit/poseidon-proof`. `scope` can now also be text.

Additionally, it include the number of inputs of the `poseidon-proof` package in the proof object, so that the `verify` function doesn't need it as a separate parameter. 

## Related Issue(s)

Closes #214

## Checklist

### DIFF
--- a/packages/poseidon-proof/src/to-bigint.ts
+++ b/packages/poseidon-proof/src/to-bigint.ts
@@ -1,0 +1,20 @@
+import type { BigNumberish } from "ethers"
+import { encodeBytes32String } from "ethers/abi"
+import { toBigInt as _toBigInt } from "ethers/utils"
+
+/**
+ * Converts a bignumberish or a text to a bigint.
+ * @param value The value to be converted to bigint.
+ * @return The value converted to bigint.
+ */
+export default function toBigInt(value: BigNumberish | Uint8Array | string): bigint {
+    try {
+        return _toBigInt(value)
+    } catch (error: any) {
+        if (typeof value === "string") {
+            return _toBigInt(encodeBytes32String(value))
+        }
+
+        throw TypeError(error.message)
+    }
+}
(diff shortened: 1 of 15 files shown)


packages/poseidon-proof/src/to-bigint.ts
```diff
@@ -0,0 +1,20 @@
+import type { BigNumberish } from "ethers"
+import { encodeBytes32String } from "ethers/abi"
+import { toBigInt as _toBigInt } from "ethers/utils"
+
+/**
+ * Converts a bignumberish or a text to a bigint.
+ * @param value The value to be converted to bigint.
+ * @return The value converted to bigint.
+ */
+export default function toBigInt(value: BigNumberish | Uint8Array | string): bigint {
```
0xjei, line 10: Q: can be a generic utils?

0xjei (approved):
amazing!

vplasencia (approved):

packages/poseidon-proof/src/to-bigint.ts
cedoor: It could be moved to utils, without `ethers` though. We'd need to replace those methods with something we already have in the utils package I guess.

vplasencia:
It looks great! Thanks. The only comment is that there seems to be an issue with the tests.

packages/poseidon-proof/src/to-bigint.ts
0xjei: makes sense - probably worth opening another issue!

PR #273, opened 2024-04-25
feat(utils): use multiple sources to download snark artifacts

sripwoud:
wip
closes privacy-scaling-explorations/snark-artifacts#65

### DIFF
--- a/packages/utils/src/snark-artifacts/config.ts
+++ b/packages/utils/src/snark-artifacts/config.ts
@@ -1,6 +1,6 @@
 import { Artifact, Proof, Version } from "../types"
 
-const CDNS = ["https://cdn.jsdelivr.net", "https://unpkg.com"]
+const CDNS = ["https://cdn.jsdelivr.net/npm", "https://unpkg.com"]
 const shuffleCdns = () => CDNS.sort(() => Math.random() - 0.5)
 const shuffleUrls = (proof: Proof, version: Version) =>
     shuffleCdns().map((cdn) => `${cdn}/@zk-kit/${proof}-artifacts@${version}/${proof}`)
--- a/packages/utils/src/snark-artifacts/config.ts
+++ b/packages/utils/src/snark-artifacts/config.ts
@@ -1,9 +1,12 @@
-import { Artifact, Proof, SnarkArtifacts, Version } from "../types"
+import { Artifact, Proof, Version } from "../types"
 
-const ARTIFACTS_BASE_URL = "https://cdn.jsdelivr.net/@zk-kit"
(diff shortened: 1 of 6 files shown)


packages/utils/src/snark-artifacts/config.ts
```diff
@@ -1,9 +1,12 @@
...
 
-const ARTIFACTS_BASE_URL = "https://unpkg.com/@zk-kit"
+const CDNS = ["https://cdn.jsdelivr.net/npm", "https://unpkg.com"]
+const shuffleCdns = () => CDNS.sort(() => Math.random() - 0.5)
+const shuffleUrls = (proof: Proof, version: Version) =>
+    shuffleCdns().map((cdn) => `${cdn}/@zk-kit/${proof}-artifacts@${version}/${proof}`)
 
 const getPackageVersions = async (proof: Proof) =>
-    fetch(`${ARTIFACTS_BASE_URL}/${proof}-artifacts`)
+    fetch(`https://registry.npmjs.org/@zk-kit/${proof}-artifacts`)
```
sripwoud, line 9: spotted a mistake there, unrelated to this PR though

sripwoud:
@sripwoud After further testing, I can confirm that unpkg is throttling downloads.
As I reimplemented the poseidon-proof to fetch all the artifacts concurrently, it would persistently fail with connection errors even before reaching the jest timeout.  
I can also confirm that with the retry fetch mechanism we can avoid that issue.  
I am happy to say that as a result (concurrent download + fetch retry to avoid connection errors), it can speed up the `poseidon-proof` tests up to around x4. (**was requiring 180s of timeout, now 45s is enough**):
https://github.com/privacy-scaling-explorations/zk-kit/actions/runs/8876339349/job/24367631912?pr=273#step:9:605  

So deciding to combine multiple CDNs with a fetch retry will both make our tests faster and the artifacts fetching more reliable.

I would wait for #275 to merge first, as both our PRs touch the same functions though.

packages/utils/src/snark-artifacts/config.ts
```diff
@@ -1,9 +1,13 @@
-import { Artifact, Proof, SnarkArtifacts, Version } from "../types"
+import { Artifact, Proof, Version } from "../types"
 
-const ARTIFACTS_BASE_URL = "https://unpkg.com/@zk-kit"
+const getBaseUrls = (proof: Proof, version: Version) => [
```
sripwoud, line 3: my initial idea was to shuffle this randomly, but as https://github.com/privacy-scaling-explorations/snark-artifacts/pull/23 told us how to rank them by download speed, we should follow that order

sripwoud:
Will reopen in here https://github.com/privacy-scaling-explorations/snark-artifacts/issues/65

PR #275, opened 2024-04-25
refactor(utils)!: make snark-artifacts utility functions dynamic

### LINKED ISSUE #272: Make functions to get SNARK artifacts more dynamic
**Description**:

The current utility functions to get the SNARK artifacts are static, meaning you need to add a new function every time the artifacts of a new project are added to the [snark-artifacts](https://github.com/privacy-scaling-explorations/snark-artifacts) repo.

They can be dynamic, i.e. a new `projectName` parameter could be added to allow devs to download the artifacts of any project is added to the repo.

cedoor:
## Description

WIP

## Related Issue(s)

Closes #272

## Checklist

-   [ ] My code follows the style guidelines of this project
-   [ ] I have performed a self-review of my code
-   [ ] My changes generate no new warnings
-   [ ] I have run `yarn style` without getting any errors


packages/eddsa-proof/src/generate.ts
```diff
@@ -1,5 +1,7 @@
 import { deriveSecretScalar } from "@zk-kit/eddsa-poseidon"
-import { SnarkArtifacts, maybeGetEdDSASnarkArtifacts, packGroth16Proof } from "@zk-kit/utils"
+import type { SnarkArtifacts } from "@zk-kit/utils"
```
sripwoud, line 2: btw if you care about enforcing and automating sorting imports (and object properties), there is 
https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/newline-after-import.md
https://github.com/mskelton/eslint-plugin-sort/tree/main

packages/poseidon-proof/src/generate.ts
```diff
@@ -30,7 +32,7 @@ export default async function generate(
 
     // allow user to override our artifacts
     // otherwise they'll be downloaded if not already in local tmp folder
-    snarkArtifacts ??= await maybeGetPoseidonSnarkArtifacts(preimages.length)
+    snarkArtifacts ??= await maybeGetSnarkArtifacts("poseidon", { parameters: [preimages.length] })
```
sripwoud, line 35: `Proof.POSEIDON`  instead of `"poseidon"`?

packages/utils/src/snark-artifacts/snark-artifacts.browser.ts
```diff
@@ -1,31 +1,21 @@
...
-function MaybeGetSnarkArtifacts(
-    proof: Proof.SEMAPHORE,
-    version?: Version
-): (treeDepth: number) => Promise<SnarkArtifacts>
-function MaybeGetSnarkArtifacts(proof: Proof, version?: Version) {
-    switch (proof) {
-        case Proof.POSEIDON:
-            return async (numberOfInputs: number) => GetSnarkArtifactUrls({ proof, numberOfInputs, version })
+export default async function maybeGetSnarkArtifacts(
+    projectName: string,
```
sripwoud, line 4: I would prefer an enum here. Eg we can use [`Proof`](https://github.com/privacy-scaling-explorations/zk-kit/blob/df57ac2f9c719fd3bc2573fc3ff612b3a634ef27/packages/utils/src/types/index.ts#L26)

packages/utils/src/snark-artifacts/snark-artifacts.browser.ts
sripwoud, line 1: Not against it. The pro is that it is simpler and less code. But I think this implementation has less robust type checking.
Because you removed the function overloads, someone could e.g try to write:
`maybeGetSnarkArtifacts("eddsa", {parameters: [123]})`
that is passing parameters that don't make sense given the projectName first parameter. In this case it has no effect, but it could turn into errors. Errors that typescript can guard against if we keep the function overloads.

packages/utils/src/snark-artifacts/snark-artifacts.browser.ts
```diff
@@ -1,31 +1,21 @@
...
+    options: {
+        parameters?: (BigNumber | number)[]
```
sripwoud, line 6: Do you use an array because you foresee having to pass more parameters to the generate function? Right now we always use 1. So why not just
`parameters?: BigNumber | number`?

packages/utils/src/snark-artifacts/snark-artifacts.node.ts
```diff
@@ -47,84 +47,24 @@ async function maybeDownload(url: string) {
...
  */
-export const maybeGetSemaphoreSnarkArtifacts = MaybeGetSnarkArtifacts(Proof.SEMAPHORE)
+export default async function maybeGetSnarkArtifacts(
+    ...pars: Parameters<typeof _maybeGetSnarkArtifacts>
+): Promise<SnarkArtifacts> {
+    const { wasm, zkey } = await _maybeGetSnarkArtifacts(...pars)
+
+    return {
+        wasm: await maybeDownload(wasm),
+        zkey: await maybeDownload(zkey)
```
sripwoud, line 68: this will fetch artifacts sequentially, instead of concurrently (if both are e.g 3MB, you wait to dl 3MB of wasm before starting 3MB of zkey).
I had observed that Concurrent download of wasm and zkey seemed to give a big download speed benefit
https://github.com/privacy-scaling-explorations/zk-kit/pull/240#discussion_r1557401950

packages/utils/src/types/index.ts
```diff
@@ -17,41 +17,18 @@ export type PackedGroth16Proof = [
...
- * {@link @zk-kit/eddsa-proof!generate | EdDSA}
- * {@link @zk-kit/poseidon-proof!generate | Poseidon}
- * {@link https://github.com/semaphore-protocol/semaphore/tree/main/packages/proof | Semaphore}.
- * @enum
- */
-export enum Proof {
-    EDDSA = "eddsa",
-    POSEIDON = "poseidon",
-    SEMAPHORE = "semaphore"
-}
```
sripwoud, line 30: don't you like the enums?
With enums I don't have to think about whether e.g I must write "eddsa" or "EdDSA" or ...

packages/utils/src/types/index.ts
```diff
@@ -17,41 +17,18 @@ export type PackedGroth16Proof = [
...
  * Semantic version.
  * @example
- * 1.0.0
+ * 1.0.0-beta
+ * 2.0.0
  * @example
  * "latest"
  */
-export type Version = `${Digit}.${Digit}.${Digit}` | "latest"
+export type Version = string | "latest"
```
sripwoud, line 34: perfect typings is not the goals, but what about

```ts
type Digit = `${number}`;
type preRelease = "alpha" | "beta";

export type Version = 
  `${Digit}.${Digit}.${Digit}` | 
  `${Digit}.${Digit}.${Digit}-${string}` |
  `${Digit}.${Digit}.${Digit}-${preRelease}.${Digit}` |
  "latest";
```

packages/utils/src/snark-artifacts/snark-artifacts.browser.ts
sripwoud: As you are joining them below, this could turn into errors.
This allows people to pass [2, 3, 4]. For functions that don't use so many arguments

packages/utils/src/snark-artifacts/snark-artifacts.node.ts
sripwoud, line 1: I agree this was lot of code we can try to reduce and maybe over engineering from my side ^^ .
There are trade offs to think about:
- if we remove the `Promise.all` stuff, we lose concurrent downloading of wasm and zkey and significant speed benefits
- if we remove the function overloads, we loose smart typechecking that match project/proof type with further parameters. This opens the door to people misusing the function especially because now we expose a more generic version

packages/poseidon-proof/tests/index.test.ts
```diff
@@ -1,85 +1,40 @@
...
     curve = await buildBn128()
-    for (const preimage of preimages) {
-        const currentPreimages = preimages.slice(0, preimage)
-        const fullProof = await generate(currentPreimages, scope)
-        const digest = computePoseidon(currentPreimages.map((preimage) => hash(preimage)))
-        proofs.push({ fullProof, digest })
-    }
-}, 180_000)
+
+    fullProof = await generate([1, 2], scope)
```
sripwoud, line 17: this will definitely be a faster test!
I know we are having lot of timeout issues atm.
I am just worried about reducing test coverage (each generate function uses different artifacts files) just to satisfy the ci...

packages/utils/src/snark-artifacts/snark-artifacts.browser.ts
```diff
@@ -1,31 +1,21 @@
...
+        version?: Version
+        cdnUrl?: string
```
sripwoud, line 8: I'd like to hear your thoughts about https://github.com/privacy-scaling-explorations/zk-kit/pull/273

cedoor:
@sripwoud I have deliberately minimized the code, but it makes sense to re-integrate parts of your code. Let me reply your comments!

packages/eddsa-proof/src/generate.ts
cedoor: Ahhh interesting! I use a nvim shortcut to sort them. I think that is also based on eslint actually.

packages/utils/src/snark-artifacts/snark-artifacts.browser.ts
cedoor: I see your point. How would you keep those checks without context about projects/proof here though?

Related: https://github.com/privacy-scaling-explorations/zk-kit/pull/275#discussion_r1580671559

packages/utils/src/snark-artifacts/snark-artifacts.browser.ts
cedoor: Arrays should also handle cases where a circuit has more than 1 parameter. So, it covers all cases where N of parameters > 0, which could be encoded as `project-1-34-<...>.wasm`. We could also use `(BigNumber | number)[] | BigNumber | number` tho.

packages/utils/src/types/index.ts
cedoor: Related: https://github.com/privacy-scaling-explorations/zk-kit/pull/275#discussion_r1580671559

packages/utils/src/types/index.ts
cedoor: This should work 👍🏽

packages/utils/src/snark-artifacts/snark-artifacts.node.ts
cedoor: > if we remove the Promise.all stuff, we lose concurrent downloading of wasm and zkey and significant speed benefits

I think we should use `Promise.all`. I don't see any downside here.

> if we remove the function overloads, we loose smart typechecking that match project/proof type with further parameters. This opens the door to people misusing the function especially because now we expose a more generic version.

I think the main question here is: do we want to keep functions as generic as possible (i.e. no need to update them based on snark-artifacts projects)? If so, we could slightly improve the checks perhaps, but most of them should probably be delegated to devs, according to their project/artifacts.

packages/poseidon-proof/tests/index.test.ts
cedoor: I've been thinking about this. I think we should ask ourselves what we are actually testing in this library.

If the artifacts are based on the circuit in `@zk-kit/circuits` (whose parameters should be tested there), they should be fine, and we don't need to test them here (a function to check the integrity should be enough). Given that, 1 couple of artifacts should be enough to test the library logic I guess.

Does it make sense? I may be missing something tho.

packages/utils/src/snark-artifacts/snark-artifacts.browser.ts
cedoor: Yess, I think that's the right way to do it! Sounds like a solid mechanism 👍🏽

packages/utils/src/snark-artifacts/snark-artifacts.browser.ts
sripwoud: I don't see how this remains possible to have type checking is we use a string as param.

packages/poseidon-proof/tests/index.test.ts
sripwoud: > @zk-kit/circuits (whose parameters should be tested there)
> we don't need to test them here
> Given that, 1 couple of artifacts should be enough to test the library logic I guess.

I agree. Testing a circuit works for different parameters is in the scope of the circuit package.

> a function to check the integrity should be enough

And we said this integrity check is in the scope of https://github.com/privacy-scaling-explorations/snark-artifacts right? see https://github.com/privacy-scaling-explorations/snark-artifacts/issues/11

packages/eddsa-proof/src/generate.ts
```diff
@@ -30,7 +32,7 @@ export default async function generate(
 
     // allow user to override our artifacts
     // otherwise they'll be downloaded if not already in local tmp folder
-    snarkArtifacts ??= await maybeGetEdDSASnarkArtifacts()
+    snarkArtifacts ??= await maybeGetSnarkArtifacts("eddsa")
```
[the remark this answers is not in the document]
cedoor: > Do you mean that there would be artifacts added to https://github.com/privacy-scaling-explorations/snark-artifacts that don't (and maybe will never) have a dedicated package in https://github.com/privacy-scaling-explorations/zk-kit (like there would be a @zk-kit/someProof-artifacts package but no @zk-kit/someProof-proof package).

Just discussed this privately. Yes, that's what I have in mind.

> I still like the idea of avoiding free text and restrict the param to valid values: those artifacts we are able to provide artifacts for.

We agreed on avoiding free text and creating an enum for each new snark-artifacts package.

> In any case we would have some work/PR to review in https://github.com/privacy-scaling-explorations/snark-artifacts no: for the PR that adds the artifact. So maybe the enum belongs to the snark-artifacts repo and any PR that adds artifacts would need to update the enum too?

We thought of having a new `@zk-kit/artifacts` package -> https://github.com/privacy-scaling-explorations/snark-artifacts/issues/11#issuecomment-2088149693. Where should it stay, zk-kit or snark-artifacts repo?

packages/poseidon-proof/tests/index.test.ts
cedoor: > And we said this integrity check is in the scope of https://github.com/privacy-scaling-explorations/snark-artifacts right? see https://github.com/privacy-scaling-explorations/snark-artifacts/issues/11

Yes!

packages/eddsa-proof/src/generate.ts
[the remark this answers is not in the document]
sripwoud: I would put it in the snark-artifacts repo. Just because I kinda "feel" that enum is more closely related to the artifacts themselves (the content of the snark artifacts repo) rather than to all the other zk-kit packages.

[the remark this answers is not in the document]
cedoor: Makes sense! I'll wrap this PR up, and then we can open a new one to move the code there.

packages/utils/src/types/index.ts
cedoor:
```ts
`${Digit}.${Digit}.${Digit}-${preRelease}.${Digit}` |
```
This doesn't really work as the previous type includes all combinations.

PR #277, opened 2024-04-26
ci: skip contract tests when PR changes are unrelated

baumstern:
This PR skips contract tests when changes are unrelated to contract files. Specifically, contract tests will only be triggered if files under `packages/**/*.sol` are modified

### DIFF
--- a/.github/workflows/pull-requests.yml
+++ b/.github/workflows/pull-requests.yml
@@ -41,6 +41,13 @@ jobs:
             - name: Checkout
               uses: actions/checkout@v4
 
+            - name: Get changed files
+              id: changed-files-contracts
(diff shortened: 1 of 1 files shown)


sripwoud (approved):
utACK

(Failing test is unrelated to this PR).

Do you think you could mirror the ci optimization you did in zk-kit in #277 and #268 in https://github.com/semaphore-protocol/semaphore?
Both are monorepos with similar setups.

baumstern:
> utACK
> 
> (Failing test is unrelated to this PR).
> 
> Do you think you could mirror the ci optimizations you did in zk-kit in #277 and #268 in https://github.com/semaphore-protocol/semaphore? Both are monorepos with similar setups.

Sure thing!

PR #292, opened 2024-05-20
chore: bump `@zk-kit/artifacts`

sripwoud:
see https://github.com/privacy-scaling-explorations/snark-artifacts/pull/57


cedoor (approved):

cedoor:
The lockfile needs to be updated I think

PR #298, opened 2024-06-08
fix(eddsa-poseidon): set default private key as buffer

### LINKED ISSUE #297: The EdDSAPoseidon class generates a private key with the wrong type
**Describe the bug**

`@zk-kit/eddsa-poseidon` provides a class (i.e. `EdDSAPoseidon`) to generate a EdDSA pub/priv key. If the constructor has no parameter, a hexadecimal private key (string) will be automatically generated. The package only supports buffers and text though, so the hexadecimal string will be treated as a text.

**Expected behavior**

The private key generated in the constructor should be a buffer.

cedoor:
## Description

The private key must always be 1 of the following supported types: buffer, uint8array, or text. The default private key generated in the constructor was a hexadecimal string and it was considered text by the internal functions. The default private key is a buffer now, which eliminates the ambiguity.

## Related Issue(s)

Fixes #297

## Checklist

### DIFF
--- a/packages/eddsa-poseidon/src/eddsa-poseidon.ts
+++ b/packages/eddsa-poseidon/src/eddsa-poseidon.ts
@@ -11,12 +11,7 @@ import {
 } from "@zk-kit/baby-jubjub"
 import type { BigNumberish } from "@zk-kit/utils"
 import { crypto, requireBuffer } from "@zk-kit/utils"
-import {
-    bigNumberishToBigInt,
-    leBigIntToBuffer,
-    leBufferToBigInt,
-    bufferToHexadecimal
-} from "@zk-kit/utils/conversions"
+import { bigNumberishToBigInt, leBigIntToBuffer, leBufferToBigInt } from "@zk-kit/utils/conversions"
 import { requireBigNumberish } from "@zk-kit/utils/error-handlers"
 import F1Field from "@zk-kit/utils/f1-field"
 import * as scalar from "@zk-kit/utils/scalar"
@@ -274,7 +269,7 @@ export class EdDSAPoseidon {
      *
      * @param privateKey The private key used for signing and public key derivation.
      */
-    constructor(privateKey: Buffer | Uint8Array | string = bufferToHexadecimal(crypto.getRandomValues(32))) {
+    constructor(privateKey: Buffer | Uint8Array | string = crypto.getRandomValues(32)) {
         this.privateKey = privateKey
         this.secretScalar = deriveSecretScalar(privateKey)
         this.publicKey = derivePublicKey(privateKey)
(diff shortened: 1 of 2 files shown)


packages/eddsa-poseidon/tests/index.test.ts
```diff
@@ -385,8 +385,8 @@ describe("EdDSAPoseidon", () => {
```
artwyman, line 385: What specific bugs or use cases are you trying to improve here?  Can you create a specific test case which demonstrates an issue so you can then demonstrate that it's fixed?

artwyman (commented):
I think this change is good, but maybe not a complete fix to the Semaphore issue which motivated it.
A few suggestions inline for the code at this level, but I'll follow up separately about the Semaphore level.

packages/eddsa-poseidon/tests/index.test.ts
```diff
@@ -385,8 +385,8 @@ describe("EdDSAPoseidon", () => {
 
         const signature = eddsa.signMessage(message)
 
-        expect(typeof eddsa.privateKey).toBe("string")
-        expect(eddsa.privateKey).toHaveLength(64)
+        expect(typeof eddsa.privateKey).toBe("object")
```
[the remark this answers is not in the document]
cedoor: Yes, it must be a string at line 378! I don't see other places to check it tho

packages/eddsa-poseidon/src/eddsa-poseidon.ts
```diff
@@ -274,7 +269,7 @@ export class EdDSAPoseidon {
      *
      * @param privateKey The private key used for signing and public key derivation.
      */
-    constructor(privateKey: Buffer | Uint8Array | string = bufferToHexadecimal(crypto.getRandomValues(32))) {
+    constructor(privateKey: Buffer | Uint8Array | string = crypto.getRandomValues(32)) {
         this.privateKey = privateKey
```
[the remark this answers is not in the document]
cedoor: >  The possible downside would be if users are surprised that the privateKey they get out isn't exactly the same as the one they put in.

That's what I'm worried about. I expect it to be the same value I put in.

packages/eddsa-poseidon/tests/index.test.ts
```diff
@@ -375,6 +375,7 @@ describe("EdDSAPoseidon", () => {
 
         const signature = eddsa.signMessage(message)
 
+        expect(typeof eddsa.privateKey).toBe("string")
```
artwyman, line 378: Since you expressed that you expect the private key to be what you put in, maybe this should be `expect(eddsa.privateKey).to.eq(privateKey)`

cedoor: Makes sense!

artwyman (approved):
LGTM at this level
I think the Semaphore level followup remains.

PR #300, opened 2024-06-11
feat(utils): add new functions to convert base64 strings

### LINKED ISSUE #299: New utility functions to convert base64 strings

Many projects could benefit from functions to convert buffers to base64 and vice versa, or text to base64 and vice versa.

`@zk-kit/utils` already provides many functions to convert buffers to other types. Since the `Buffer` library already includes methods to convert base64 strings to other values, it's quite easy to support the conversions mentioned above.

cedoor:
## Description

This PR adds 4 functions to convert base64 strings from buffers and text.

## Related Issue(s)

Closes #299

## Checklist

### DIFF
--- a/packages/utils/src/conversions.ts
+++ b/packages/utils/src/conversions.ts
@@ -222,8 +222,8 @@ export function bufferToHexadecimal(value: Buffer | Uint8Array): string {
 }
 
 /**
- * Converts a buffer to a base64 string. It accepts 'Buffer' or 'Uint8Array'.
- * @param value The buffer to convert.
+ * Converts bytes to a base64 string. It accepts 'Buffer' or 'Uint8Array'.
+ * @param value The bytes to convert.
  * @returns The converted base64 string.
  */
 export function bufferToBase64(value: Buffer | Uint8Array): string {
@@ -233,7 +233,7 @@ export function bufferToBase64(value: Buffer | Uint8Array): string {
 }
 
 /**
- * Converts a base64 string to a buffer. This function does not check
+ * Converts a base64 string to bytes (i.e. a buffer). This function does not check
  * if the input value is a valid base64 string.
  * @param value The base64 string to convert.
  * @returns The converted buffer.
--- a/packages/utils/src/conversions.ts
+++ b/packages/utils/src/conversions.ts
@@ -234,7 +234,8 @@ export function bufferToBase64(value: Buffer | Uint8Array): string {
 
 /**
  * Converts a base64 string to bytes (i.e. a buffer). This function does not check
- * if the input value is a valid base64 string.
+ * if the input value is a valid base64 string. If there are unsupported characters
+ * they will be ignored.
  * @param value The base64 string to convert.
  * @returns The converted buffer.
  */
@@ -257,7 +258,8 @@ export function textToBase64(value: string): string {
 
 /**
  * Converts a base64 string to text (utf8). This function does not check
- * if the input value is a valid base64 string.
+ * if the input value is a valid base64 string. If there are unsupported characters
+ * they could be ignored and the result may be unexpected.
  * @param value The base64 string to convert.
  * @returns The converted text.
  */
--- a/packages/utils/src/conversions.ts
+++ b/packages/utils/src/conversions.ts
@@ -13,7 +13,7 @@
  */
 
 import { Buffer } from "buffer"
-import { requireBigInt, requireBigNumberish, requireHexadecimal, requireTypes } from "./error-handlers"
+import { requireBigInt, requireBigNumberish, requireHexadecimal, requireString, requireTypes } from "./error-handlers"
 import { isBuffer, isHexadecimal, isUint8Array } from "./type-checks"
 import { BigNumber, BigNumberish } from "./types"
 
@@ -219,4 +219,50 @@ export function bufferToHexadecimal(value: Buffer | Uint8Array): string {
     requireTypes(value, "value", ["Buffer", "Uint8Array"])
 
     return Buffer.from(value).toString("hex")
+}
+
+/**
+ * Converts a buffer to a base64 string. It accepts 'Buffer' or 'Uint8Array'.
+ * @param value The buffer to convert.
+ * @returns The converted base64 string.
+ */
+export function bufferToBase64(value: Buffer | Uint8Array): string {
+    requireTypes(value, "value", ["Buffer", "Uint8Array"])
+
+    return Buffer.from(value).toString("base64")
+}
+
+/**
+ * Converts a base64 string to a buffer. This function does not check
+ * if the input value is a valid base64 string.
+ * @param value The base64 string to convert.
+ * @returns The converted buffer.
+ */
+export function base64ToBuffer(value: string): Buffer {
+    requireString(value, "value")
+
+    return Buffer.from(value, "base64")
+}
+
+/**
+ * Converts text (utf8) to a base64 string.
+ * @param value The text to convert.
+ * @returns The converted base64 string.
+ */
+export function textToBase64(value: string): string {
+    requireString(value, "value")
+
+    return Buffer.from(value, "utf8").toString("base64")
+}
+
+/**
+ * Converts a base64 string to text (utf8). This function does not check
+ * if the input value is a valid base64 string.
+ * @param value The base64 string to convert.
+ * @returns The converted text.
+ */
+export function base64ToText(value: string): string {
+    requireString(value, "value")
+
+    return Buffer.from(value, "base64").toString("utf8")
 }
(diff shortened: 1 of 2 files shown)


packages/utils/src/conversions.ts
```diff
@@ -220,3 +220,49 @@ export function bufferToHexadecimal(value: Buffer | Uint8Array): string {
...
+ * @returns The converted base64 string.
+ */
+export function bufferToBase64(value: Buffer | Uint8Array): string {
+    requireTypes(value, "value", ["Buffer", "Uint8Array"])
+
+    return Buffer.from(value).toString("base64")
+}
+
+/**
+ * Converts a base64 string to a buffer. This function does not check
```
artwyman, line 236: Can you document (and ideally also test in the utest) what happens if it's not?
A quick experiment in Node suggests that what happens is that invalid characters are ignored, rather than an error being thrown.

packages/utils/src/conversions.ts
```diff
@@ -220,3 +220,49 @@ export function bufferToHexadecimal(value: Buffer | Uint8Array): string {
...
+ */
+export function textToBase64(value: string): string {
+    requireString(value, "value")
+
+    return Buffer.from(value, "utf8").toString("base64")
+}
+
+/**
+ * Converts a base64 string to text (utf8). This function does not check
+ * if the input value is a valid base64 string.
```
artwyman, line 260: As above, document what happens to invalid inputs.
In this case there are two different encodings which could go wrong.

packages/utils/src/conversions.ts
```diff
@@ -220,3 +220,49 @@ export function bufferToHexadecimal(value: Buffer | Uint8Array): string {
...
+
+/**
+ * Converts text (utf8) to a base64 string.
+ * @param value The text to convert.
+ * @returns The converted base64 string.
+ */
+export function textToBase64(value: string): string {
+    requireString(value, "value")
+
+    return Buffer.from(value, "utf8").toString("base64")
```
artwyman, line 255: I'm not confident enough in my knowledge of Unicode in JavaScript to be sure this is the right way to deal with non-ASCII text.  I'd suggest some unit tests which exercise this with some non-ASCII text, including multilingual stuff (accents, cyrilic, Chinese) and some emojis, and see that they round-trip correctly through `textToBase64` -> `base64ToText`.

packages/utils/src/conversions.ts
```diff
@@ -220,3 +220,49 @@ export function bufferToHexadecimal(value: Buffer | Uint8Array): string {
 
     return Buffer.from(value).toString("hex")
 }
+
+/**
+ * Converts a buffer to a base64 string. It accepts 'Buffer' or 'Uint8Array'.
+ * @param value The buffer to convert.
+ * @returns The converted base64 string.
+ */
+export function bufferToBase64(value: Buffer | Uint8Array): string {
```
artwyman, line 229: "bytes" instead of "buffer" maybe?  Since the use of the Buffer type is optional for this function and its inverse.

packages/utils/tests/conversions.test.ts
```diff
@@ -170,6 +174,40 @@ describe("Conversions", () => {
         })
     })
 
+    describe("# base64ToBuffer / bufferToBase64", () => {
+        const testBase64 = "SGVsbG8sIFdvcmxkIQ=="
```
artwyman, line 178: Also test some arbitrary byte data which isn't text?  You want this to be able to handle things like null bytes, control characters, etc.

packages/utils/tests/conversions.test.ts
```diff
@@ -170,6 +174,40 @@ describe("Conversions", () => {
...
+
+        it("Should convert a buffer to a base64 string", async () => {
+            const result = bufferToBase64(Buffer.from(testText))
+
+            expect(result).toBe(testBase64)
+        })
+    })
+
+    describe("# base64ToText / textToBase64", () => {
+        const testBase64 = "SGVsbG8sIFdvcmxkIQ=="
```
artwyman, line 195: As I mention above, more test cases using non-ASCII text would be a good idea.

artwyman (changes requested):
Interface and functionality looks good (with the caveat that I may not be sure about the Unicode stuff being right).  My suggestions are about documentation and testing.

packages/utils/src/conversions.ts
cedoor: I confirm it ignores invalid characters. The documentation also mentions it: https://nodejs.org/api/buffer.html#buffers-and-character-encodings. Checking if a base64 is valid is not trivial and I'd avoid that in this function. I'll add more documentation.

packages/utils/src/conversions.ts
cedoor: In this case `base64ToText` -> `textToBase64` gives a result that is different from the input. So I'll also mention the result may be unexpected.

packages/utils/src/conversions.ts
cedoor: `utf-8` supports all characters in the Unicode standard and it's the preferred encoding for web content. I tested many different characters and they all work 👍🏽

cedoor:
> Interface and functionality looks good (with the caveat that I may not be sure about the Unicode stuff being right). My suggestions are about documentation and testing.

Please, take a look again and let me know if the documentation and tests I added are ok.

packages/utils/src/conversions.ts
artwyman: I think that's fine so long as the behavior is documented.

A regex can check validity of Base64 pretty reasonably, but that's non-trivial at the level of cost of Base64 conversion, yes.  For PODs I used a regex at the level of key/signature decoding, when I know the specific size and other properties I'm looking for.

packages/utils/src/conversions.ts
artwyman: If you're going to refer to bytes not buffer shouldn't that be reflected in the function name, not just the comments?  I.e. consider renaming the functions to byteToBase64 and base64ToBytes?

packages/utils/src/conversions.ts
artwyman: Agreed, I just wasn't sure what the semantics of JavaScript strings were in that regard.  In some languages, strings separate their internal encoding from external encodings.

One case to consider documenting here is what happens to invalid utf8 sequences in the input.  I presume they'll probably get ignored similarly to the invalid base64 in other cases.

artwyman (approved):
A few more suggestions in comments, but they're optional so you can decide whether or not to make further changes.  Functionally this seems good.

packages/utils/src/conversions.ts
cedoor: Can you make an example? An input like "��" will be correctly converted.

packages/utils/src/conversions.ts
cedoor: Well, we should update many names in this library in that case. I actually think we can consider Uint8Array to be a type of buffer in this context no?

cedoor:
> A few more suggestions in comments, but they're optional so you can decide whether or not to make further changes. Functionally this seems good.

Thank you 🙏🏽  I'll merge this PR. We can open a new one if necessary.

packages/utils/src/conversions.ts
artwyman: Funny.  I don't know if you really meant to send question-mark charaters or if GitHub + Chrome isn't decoding something.
image

I was thinking of strings which contain non-UTF data.  Not every possible sequence of bytes is valid UTF8.  Look at the encding [here](https://en.wikipedia.org/wiki/UTF-8#Encoding) and you can see if the first byte indicates a multi-byte sequence, then the remaining bytes must have high bits 10, and if they don't it would be an invalid encoding.

It may not be worth worrying about, as bad utf8 is much less of a concern than bad base64.  The structure of the code simply makes me think about any decoding being subject to failure.  I think it's fine that the answer here is "we do what JavaScript does" or "we do what Buffer does".  The main bit of clarification I'd be looking for as a user would be whether the result is an exception, or the result is truncated output.

packages/utils/src/conversions.ts
artwyman: At the code level I think it's the opposite?  Buffer is a type of Uint8Array.

If you mean "buffer" as an english word, then yes, a Uint8Array is a way of representing a "buffer" of bytes, but I'd say that's a confusing definition to rely on when there's a class called Buffer involved.

This feels like bikeshedding, though, and I don't really care enough for it to be worth arguing.

packages/utils/src/conversions.ts
cedoor: I actually sent two question marks, but that's a character that should be correctly supported actually right? :D

packages/utils/src/conversions.ts
cedoor: I think it makes sense, as "Buffer" in the function name may suggest it's a "Buffer" type. I'd keep this small inaccuracy tho since we would need to change many names to keep it consistent and it would be a breaking change.

PR #309, opened 2024-07-25
chore(poseidon-proof): pin artifacts version

### LINKED ISSUE #308: pin artifacts deps
The version of the ZK artifacts to download when generating a proof needs to be specified, as the latest version will be used by default, and when a new version is released the old version would still point to the latest, which would no longer be the right version.dd any other context about the problem here.

sripwoud:
Fix #308
Mirror of https://github.com/semaphore-protocol/semaphore/pull/826/commits/9f6516c060de803545bcaa7e415cabaafaa81527

### DIFF
--- a/packages/poseidon-proof/src/generate.ts
+++ b/packages/poseidon-proof/src/generate.ts
@@ -31,7 +31,10 @@ export default async function generate(
 
     // allow user to override our artifacts
     // otherwise they'll be downloaded if not already in local tmp folder
-    snarkArtifacts ??= await maybeGetSnarkArtifacts(Project.POSEIDON, { parameters: [preimages.length] })
+    snarkArtifacts ??= await maybeGetSnarkArtifacts(Project.POSEIDON, {
+        parameters: [preimages.length],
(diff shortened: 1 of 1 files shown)


packages/poseidon-proof/src/generate.ts
```diff
@@ -31,7 +31,10 @@ export default async function generate(
 
     // allow user to override our artifacts
     // otherwise they'll be downloaded if not already in local tmp folder
-    snarkArtifacts ??= await maybeGetSnarkArtifacts(Project.POSEIDON, { parameters: [preimages.length] })
+    snarkArtifacts ??= await maybeGetSnarkArtifacts(Project.POSEIDON, {
+        parameters: [preimages.length],
+        version: "1.0.0-beta.1"
```
sripwoud, line 36: matches https://github.com/privacy-scaling-explorations/snark-artifacts/blob/41408fc596678ef7dfb18adf0f23be123e03ea0c/packages/poseidon/package.json#L5

PR #313, opened 2024-08-07
refactor(lean-imt)!: update function to import exported trees

### LINKED ISSUE #312: Improvements in the LeanIMT `import` function

In order to use the `import` function of the LeanIMT, an empty tree must be initialized. Then the import function can be used to get the nodes exported previously. This two steps could be one:

```diff
- const tree = new LeanIMT(hash)
- tree.import(nodes)
+ const tree = LeanIMT.import(hash, nodes) 
```

Also, the import function should take a callback to convert node types. The exported nodes are strings, but the default type used for the LeanIMT is `bigint`. Devs should explicitly convert the node types to whatever they need (strings, bigints etc).

```ts
const tree = LeanIMT.import(hash, nodes, (_, v) => (typeof v === "string" ? BigInt(v) : v))
```

**Additional context**

Originated by a bug in the Semaphore `@semaphore-protocol/group` lib: https://github.com/semaphore-protocol/semaphore/issues/847.

cedoor:
## Description

The `import` function was made static and a parameter was added to convert node types.

## Related Issue(s)

Closes #312

## Other information

Please notice that this PR introduces a breaking change.

## Checklist

### DIFF
--- a/packages/lean-imt/src/lean-imt.ts
+++ b/packages/lean-imt/src/lean-imt.ts
@@ -329,31 +329,33 @@ export default class LeanIMT<N = bigint> {
     /**
      * It imports an entire tree by initializing the nodes without calculating
      * any hashes. Note that it is crucial to ensure the integrity of the tree
-     * before or after importing it.
+     * before or after importing it. If the reviver function is not defined, node
+     * values will be converted from strings to bigints by default.
      * @param hash The hash function used to create nodes.
      * @param nodes The stringified JSON of the tree.
-     * @param reviver A function that transforms the results. This function is called for each node of the tree and can
-     * be used to convert the node types.
+     * @param map A function to map each node of the tree and convert their types.
      * @returns A LeanIMT instance.
      */
-    static import<N = bigint>(
-        hash: LeanIMTHashFunction<N>,
-        nodes: string,
-        reviver?: (this: any, key: string, value: any) => any
-    ): LeanIMT<N> {
+    static import<N = bigint>(hash: LeanIMTHashFunction<N>, nodes: string, map?: (value: string) => N): LeanIMT<N> {
         requireDefined(hash, "hash")
         requireDefined(nodes, "nodes")
         requireFunction(hash, "hash")
         requireString(nodes, "nodes")
 
-        if (reviver) {
-            requireDefined(reviver, "reviver")
-            requireFunction(reviver, "reviver")
+        if (map) {
+            requireDefined(map, "map")
+            requireFunction(map, "map")
         }
 
         const tree = new LeanIMT<N>(hash)
 
-        tree._nodes = JSON.parse(nodes, reviver ?? ((_, v) => (typeof v === "string" ? BigInt(v) : v)))
+        tree._nodes = JSON.parse(nodes, (_, value) => {
+            if (typeof value === "string") {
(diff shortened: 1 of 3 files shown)


vplasencia (approved):

packages/lean-imt/src/lean-imt.ts
```diff
@@ -330,17 +330,31 @@ export default class LeanIMT<N = bigint> {
      * It imports an entire tree by initializing the nodes without calculating
      * any hashes. Note that it is crucial to ensure the integrity of the tree
      * before or after importing it.
-     * The tree must be empty before importing.
+     * @param hash The hash function used to create nodes.
      * @param nodes The stringified JSON of the tree.
+     * @param reviver A function that transforms the results. This function is called for each node of the tree and can
```
artwyman, line 335: Maybe point out that this is as defined by JSON.parse, so that's where to look for more details on how this works?
Without that dependency, I would've been asking why the reviver isn't more strongly typed.

packages/lean-imt/src/lean-imt.ts
```diff
@@ -330,17 +330,31 @@ export default class LeanIMT<N = bigint> {
...
+     * be used to convert the node types.
+     * @returns A LeanIMT instance.
      */
-    public import(nodes: string) {
+    static import<N = string>(
+        hash: LeanIMTHashFunction<N>,
+        nodes: string,
+        reviver?: (this: any, key: string, value: any) => any
```
artwyman, line 342: There's an asymmetry that export() automatically handles conversion of bigints to strings, while import() requires the user to explicitly write the code to do that.  Seems like there should be a default behavior so that export() followed by import() can work in a default way.  Put a different way, I'd expect when using bigint that I'd either have to specify a conversion for both import and export, or for both, but not for only one of them.

packages/lean-imt/tests/lean-imt.test.ts
```diff
@@ -335,54 +335,43 @@ describe("Lean IMT", () => {
         it("Should export a tree", () => {
             const tree = new LeanIMT(poseidon, leaves)
 
-            const exportedTree = tree.export()
+            const nodes = tree.export()
```
artwyman, line 338: Could use some tests of trees with types other than bigint, and with different reviver funtions.

artwyman (approved):
This looks reasonable to me.  Importing being static seems clearer than being an instance method which overwrites the state.
I've made some suggestions inline about the type conversion and reviver functionality, though this PR in't changing the behavior there, so I'd call them optional.
Note that my use of LeanIMT in POD/Zupass doesn't include export/import functionality, so I won't be directly affected by this at upgrade time.

packages/lean-imt/src/lean-imt.ts
cedoor: True, N is `bigint` by default in the `import` function too now. And there's a conversion from strings to bigints if the reviver function is not defined.

packages/lean-imt/src/lean-imt.ts
cedoor: In the end I made that function much simpler. Can you take a look again?

```ts
// Convert node types to numbers.
LeanIMT.import<number>(hash, nodes, Number) 
```

Issue #230, opened 2024-03-23
Restrict types of `message` in eddsa-poseidon's `signMessage`


artwyman:
**Describe the improvement you're thinking about**

This issue is conceptually similar to #188, now applied to the `message` argument to eddsa-poseidon's `signMessage` rather than the private key.  This argument is typed as `BigNumberish` and the docs don't say anything special about how it's interpreted.  The message of the underlying Poseidon algorithm is expected to be a single field element (bigint mod r).

The `checkMessage` function used to check this input allows arbitrary strings.  If a string is valid decimal or hex, it'll be converted to a bigint as digits.  Otherwise it'll be converted to a Buffer as raw bytes.  This works so long as the string isn't more than 32 characters.  If it's longer, it'll trigger an exception later from `leBigIntToBuffer` which uses a size argument.

This might be intended to be a convenience feature for signing string messages.  If so, it would need to be more clearly documented.  I think the size limitation and the potential for confusion about whether a small string should be treated as digits or bytes suggests it would be better to remove this case and reject string inputs which can't be interpreted as stringified bigints.

**Additional context**
I came across this when writing tests for handling of bad inputs in the Zupass repo.  I passed a string expecting a TypeError and was surprised to get no exception.  I've added my own type check to avoid the potential confusion here, so I'm not blocked by this issue.

f3r10:
hi @cedoor could I take this issue? I am part of the PSE 2024 core program

cedoor:
@f3r10 sure! Is there enough information?

cedoor:
> @cedoor the only type supported for the message parameter should be `bitint` or `bigint | number | Buffer | Uint8Array`?

`message` is limited to 32 bytes and is converted to `bigint` internally, so one simple option (which would make it more consistent with the other types) would be to have `BigNumber` imo. In case we need to add documentation to explain how devs can convert text or buffers to bigint tho (should happen rarely).

Tagging @artwyman here. Wdyt?

f3r10:
oh ok, using a BigNumber I created an additional unit test with
```
        const message = Buffer.from(crypto.getRandomValues(31))
``` 
with 31 as an example, the test passes but with 34 the test fails. Would be that ok?

cedoor:
> Would be that ok?

Yes, please make sure to update/add tests to check the correct type. `BigNumber` is either `bigint` or `string` (`string` must still be a stringified bigint).

f3r10:
[...]

PR #318, opened 2024-09-07
refactor(eddsa-poseidon)!: restrict message types

f3r10:
BREAKING CHANGE: message type

re #230

## Description
The type supported for the EdDSA message has been reduced to BigNumber (either bigint or string which  must still be a stringified bigint))

## Checklist

### DIFF
--- a/packages/eddsa-poseidon/src/eddsa-poseidon.ts
+++ b/packages/eddsa-poseidon/src/eddsa-poseidon.ts
@@ -9,7 +9,7 @@ import {
     subOrder,
     unpackPoint
 } from "@zk-kit/baby-jubjub"
-import type { BigNumberish } from "@zk-kit/utils"
+import type { BigNumber, BigNumberish } from "@zk-kit/utils"
 import { crypto, requireBuffer } from "@zk-kit/utils"
 import { bigNumberishToBigInt, leBigIntToBuffer, leBufferToBigInt } from "@zk-kit/utils/conversions"
 import { requireBigNumberish } from "@zk-kit/utils/error-handlers"
@@ -87,7 +87,7 @@ export function derivePublicKey(privateKey: Buffer | Uint8Array | string): Point
  * @param message The message to be signed.
  * @returns The signature object, containing properties relevant to EdDSA signatures, such as 'R8' and 'S' values.
  */
-export function signMessage(privateKey: Buffer | Uint8Array | string, message: BigNumberish): Signature<bigint> {
+export function signMessage(privateKey: Buffer | Uint8Array | string, message: BigNumber): Signature<bigint> {
     // Convert the private key to buffer.
     privateKey = checkPrivateKey(privateKey)
 
@@ -121,7 +121,7 @@ export function signMessage(privateKey: Buffer | Uint8Array | string, message: B
  * @param publicKey The public key associated with the private key used to sign the message.
  * @returns Returns true if the signature is valid and corresponds to the message and public key, false otherwise.
  */
-export function verifySignature(message: BigNumberish, signature: Signature, publicKey: Point): boolean {
+export function verifySignature(message: BigNumber, signature: Signature, publicKey: Point): boolean {
     if (
         !isPoint(publicKey) ||
         !isSignature(signature) ||
@@ -281,7 +281,7 @@ export class EdDSAPoseidon {
      * @param message The message to be signed.
      * @returns The signature of the message.
      */
-    signMessage(message: BigNumberish): Signature<bigint> {
+    signMessage(message: BigNumber): Signature<bigint> {
         return signMessage(this.privateKey, message)
     }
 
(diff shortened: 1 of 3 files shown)


cedoor (commented):
Thanks @f3r10 for this PR.

Looks like you need to fix the error message in your test.

packages/eddsa-poseidon/src/utils.ts
```diff
@@ -1,5 +1,5 @@
 import { Point } from "@zk-kit/baby-jubjub"
-import type { BigNumberish } from "@zk-kit/utils"
+import { type BigNumberish } from "@zk-kit/utils"
```
artwyman, line 2: I honestly don't know what this `type` keyword is for.  The build works fine for me locally if I simply remove it entirely, so I suggest doing that if it passes the GitHub checks.

packages/eddsa-poseidon/src/eddsa-poseidon.ts
```diff
@@ -87,7 +87,7 @@ export function derivePublicKey(privateKey: Buffer | Uint8Array | string): Point
  * @param message The message to be signed.
  * @returns The signature object, containing properties relevant to EdDSA signatures, such as 'R8' and 'S' values.
  */
-export function signMessage(privateKey: Buffer | Uint8Array | string, message: BigNumberish): Signature<bigint> {
+export function signMessage(privateKey: Buffer | Uint8Array | string, message: BigNumber): Signature<bigint> {
```
artwyman, line 90: I believe the intent here is to change not only the types but the handling and checking of the inputs here.  A BigNumber is limited to bigint or a stringified bigint (not an arbitrary string, not a buffer).  The checkMessage function is the place where you should be checking that and converting to bigint appropriately.

packages/eddsa-poseidon/tests/index.test.ts
```diff
@@ -96,7 +96,7 @@ describe("EdDSAPoseidon", () => {
     it("Should sign a message (hexadecimal)", async () => {
         const message = "0x12"
 
-        const signature = signMessage(privateKey, message)
+        const signature = signMessage(privateKey, BigInt(message))
```
artwyman, line 99: This is a stringified bigint, so should be accepted as a BigNumber without explicit conversion.

I also suggest adding another test case with a decimal string like "123" which should also work.  Anything parsable by BigInt(s) should work.

packages/eddsa-poseidon/tests/index.test.ts
```diff
@@ -117,7 +117,7 @@ describe("EdDSAPoseidon", () => {
         expect(signature.S).toBe(circomlibSignature.S)
     })
 
-    it("Should sign a message (string)", async () => {
+    it("Should sign a message if less than 32 bytes (string)", async () => {
```
artwyman, line 120: I think this case shouldn't be supported without an explicit conversion to bigint

packages/eddsa-poseidon/tests/index.test.ts
```diff
@@ -129,6 +129,13 @@ describe("EdDSAPoseidon", () => {
         expect(signature.S).toBe(circomlibSignature.S)
     })
 
+    it("Should fail if message is larger than 32 bytes (string)", async () => {
+        const message = bufferToBigInt(Buffer.from(crypto.getRandomValues(34)))
+
+        const fun = () => signMessage(privateKey, message)
+        expect(fun).toThrow("Size 32 is too small, need at least 33 bytes")
```
artwyman, line 136: This test is failing on GitHub

packages/eddsa-poseidon/src/eddsa-poseidon.ts
artwyman: Also per @cedoor's comment on https://github.com/privacy-scaling-explorations/zk-kit/issues/230 there should be some documentation above about the fact that the message is a single BabyJub field element, represented as a bigint.  Best-practice for users with other types of input is probably to hash it (e.g. with a poseidon hash), or encode their inputs bytes directly if small enough (e.g. using bufferToBigInt()).

artwyman (changes requested):
I think this change needs to include functional code not just types and tests.  In particular the checkMessage function is where the input needs to be validated and converted appropriately.

packages/eddsa-poseidon/src/utils.ts
cedoor: Imports could be values or types. TS can infer what kind of import that is but the `type` keyword here makes it more explicit.

f3r10:
awesome!! thanks @artwyman for your feedback. I am going to work through your comments right away.

cedoor:
Hi @f3r10, are you still going to work on this?

f3r10:
My apologies, @cedoor. I started a new job and have been really busy. I hope to contribute once I settle into it.

PR #320, opened 2024-09-11
fix: add dts plugin to roll-up definition files

### LINKED ISSUE #319: Non-ESM-compliant exports break for certain downstream consumers
**Describe the bug**
Modern ESM-compliant code requires that all `import` statements give a path to the file being imported including a file extension (typically `.js`). This is because ESM modules should be importable in a browser context, e.g. `import "./path/to/import/index.js"` maps to HTTP `GET /path/to/import/index.js`. Unlike filesystem-based imports, browsers cannot cheaply try multiple file extensions in order to locate the file, so full file extensions are mandatory in ESM imports.

However, older TypeScript code which uses `moduleResolution: "bundler"` often contains imports without filenames, e.g. `import Component from "./Component"` rather than `import Component from "/.Component.js"`. This affects both JavaScript output and TypeScript type definitions (`.d.ts` files).

With the newer `NodeNext` module resolution strategy, TypeScript will not allow non-standard imports which do not have the file extension. Importing from a package with "incorrect" imports will break. This applies to both JavaScript and TypeScript definition imports.

If a downstream project is an ESM module (`"type": "module"` in `package.json`) and is using `"moduleResolution": "NodeNext"` in `tsconfig.json`, then imports from ZK-kit packages will not work properly, even if those packages could be imported successfully with `"moduleResolution": "bundler"`. This is because the TypeScript definition files have non-ESM-compliant imports, e.g.:
```
export * from "./eddsa-poseidon";
export * from "./types";
```

**To Reproduce**
There is a reproduction of the issue here: https://github.com/robknight/esm-import-repro
Clone the repo and run `pnpm build`, and you should see this error message:
```
src/index.ts:1:10 - error TS2305: Module '"@zk-kit/eddsa-poseidon"' has no exported member 'derivePublicKey'.

1 import { derivePublicKey } from "@zk-kit/eddsa-poseidon";
           ~~~~~~~~~~~~~~~
```

**Expected behavior**
It is clearly intended that this function (and others!) should be exported. However, TypeScript cannot follow the imports from `dist/types/index.d.ts` because they lack file extensions.

I could switch my project to use a bundler, and `"moduleResolution": "bundler"`, but since I want to create a package which can be consumed by other people, I would simply be pushing the problem to them: they would also be forced to use `"moduleResolution": "bundler"` and would be unable to use `"moduleResolution": "NodeNext"` when importing my package.

One fix is to include file extensions in the TypeScript source, e.g. `export * from "./eddsa-poseidon.js"; export * from "./types/index.js";`. Note that the file extension should be `.js` even though the source file is `.ts`.

In this specific case, another option would be to bundle all of your TypeScript definitions into a single output file. The rollup `dts` plugin can do this. Because there's only one file, there are no `import` statements, and so they can't be non-compliant.

**Screenshots**

**Technologies (please complete the following information):**

- Node v20
- TypeScript 5.5

**Additional context**
https://nodejs.org/api/esm.html#esm_packages
https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c
https://www.reddit.com/r/typescript/comments/1b87o96/esm_on_nodejs_file_extension_mandatory/

cedoor:
## Description

This PR integrates [`rollup-plugin-dts`](https://github.com/Swatinem/rollup-plugin-dts) to roll-up definition files for each package and fix https://github.com/privacy-scaling-explorations/zk-kit/issues/319.

## Related Issue(s)

Fixes #319

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,6 +62,7 @@
         "lint-staged": "^15.2.2",
         "prettier": "^3.2.5",
         "rimraf": "^5.0.5",
+        "rollup-plugin-dts": "^6.1.1",
         "ts-jest": "^29.1.2",
         "ts-node": "^10.9.2",
         "tslib": "^2.6.2",
--- a/packages/baby-jubjub/build.tsconfig.json
+++ b/packages/baby-jubjub/build.tsconfig.json
@@ -1,8 +1,7 @@
 {
     "extends": "../../tsconfig.json",
     "compilerOptions": {
-        "baseUrl": ".",
-        "declarationDir": "dist/types"
+        "baseUrl": "."
     },
     "include": ["src"]
 }
--- a/packages/baby-jubjub/package.json
+++ b/packages/baby-jubjub/package.json
@@ -8,10 +8,10 @@
     "unpkg": "dist/index.iife.min.js",
     "jsdelivr": "dist/index.iife.min.js",
     "main": "dist/index.js",
-    "types": "dist/types/index.d.ts",
+    "types": "dist/index.d.ts",
     "exports": {
         ".": {
-            "types": "./dist/types/index.d.ts",
+            "types": "./dist/index.d.ts",
             "require": "./dist/index.cjs",
             "default": "./dist/index.js"
         }
@@ -43,6 +43,7 @@
         "rimraf": "^5.0.5",
         "rollup": "^4.12.0",
         "rollup-plugin-cleanup": "^3.2.1",
+        "rollup-plugin-dts": "^6.1.1",
         "rollup-plugin-polyfill-node": "^0.13.0"
     },
     "dependencies": {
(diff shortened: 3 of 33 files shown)


sripwoud (commented):
~~I forked the repro repo mentioned in the issue to test importing all packages.
I can import them all successfully except `@zk-kit/baby-jujub`
I don't understand why, I don't see anything different in the config of that package~~

~~See https://github.com/sripwoud/esm-import-repro/pull/1~~

sripwoud (approved):
tACK
see https://github.com/sripwoud/esm-import-repro/pull/1

vplasencia (approved):

robknight:
Works for me!

Issue #324, opened 2024-09-11
`mulPointScalar` vulnerable to timing attacks


ChinoCribioli:
The `mulPointScalar` method is implemented with the regular 'square and multiply' algorithm, which is prone to [timing attacks](https://en.wikipedia.org/wiki/Timing_attack#Examples) due to the fact that the number of EC point additions depends on the number of 1's in the binary expression of the scalar. To fix this it is necessary to implement an algorithm to multiply a point by an integer in constant time.

PR #325, opened 2024-09-11
Constant-time EC point multiplication (Montgomery ladder) implementation

ChinoCribioli:
## Description

The `mulPointScalar` method is implemented with the regular [square and multiply algorithm](https://en.wikipedia.org/wiki/Exponentiation_by_squaring), which is prone to [timing attacks](https://en.wikipedia.org/wiki/Timing_attack#Examples) due to the fact that the number of EC point additions depends on the number of 1's in the binary expression of the scalar.

With this implementation (called [Montgomery Ladder](https://en.wikipedia.org/wiki/Exponentiation_by_squaring#Montgomery's_ladder_technique)) this is avoided by maintaining the number of additions depending only on the length of the binary representation of the scalar.

As documented in my implementation of the method, the algorithm works because of the following invariant: At each step, `R0` will be `r_0*base` where `r_0` is the prefix of `e` written in binary and `R1` will be `(r_0+1)*base`. In other words: at iteration `i` of the loop, `r_0`'s binary representation will be the first `i+1` most significant bits of `e`. If the upcoming bit is a 0, we just have to double `R0` and add `R0` to `R1` to maintain the invariant. If it is a 1, we have to double `R0` and add `1*base` (or add `R1`, which is the same as `(r_0+1)*base`), and double `R1` to maintain the invariant.

## Related Issue(s)

This is an intended fix for [issue #324](https://github.com/privacy-scaling-explorations/zk-kit/issues/324).

## Other information

Aside from the new implementation of the `mulPointScalar` method, I added some tests to automatically check basic behaviors such as EC point additions, EC point multiplications, and the orders of the generating/base points used in the protocol.

These tests are independent of the new implementation of the multiply method and are intended to make the test base more robust.

## Checklist

### DIFF
--- a/packages/baby-jubjub/src/baby-jubjub.ts
+++ b/packages/baby-jubjub/src/baby-jubjub.ts
@@ -70,20 +70,31 @@ export function addPoint(p1: Point<bigint>, p2: Point<bigint>): Point<bigint> {
  * @returns The resulting point representing the public key.
  */
 export function mulPointEscalar(base: Point<bigint>, e: bigint): Point<bigint> {
-    let res: Point<bigint> = [Fr.e(BigInt(0)), Fr.e(BigInt(1))]
-    let rem: bigint = e
-    let exp: Point<bigint> = base
-
-    while (!scalar.isZero(rem)) {
-        if (scalar.isOdd(rem)) {
-            res = addPoint(res, exp)
+    console.log(e)
+    const eBits: Array<boolean> = []
+    while (!scalar.isZero(e)) {
+        if (scalar.isOdd(e)) {
+            eBits.push(true)
+        } else {
+            eBits.push(false)
         }
-
-        exp = addPoint(exp, exp)
-        rem = scalar.shiftRight(rem, BigInt(1))
+        e = scalar.shiftRight(e, BigInt(1))
     }
 
-    return res
+    let R0: Point<bigint> = base
+    let R1: Point<bigint> = addPoint(base, base)
+
+    for (const bit of eBits.slice(0, -1).reverse()) {
+        if (bit) {
+            R0 = addPoint(R0, R1)
+            R1 = addPoint(R1, R1)
+        } else {
+            R1 = addPoint(R0, R1)
+            R0 = addPoint(R0, R0)
+        }
+    }
+
+    return R0
 }
 
 /**
--- a/packages/baby-jubjub/src/baby-jubjub.ts
+++ b/packages/baby-jubjub/src/baby-jubjub.ts
@@ -65,6 +65,11 @@ export function addPoint(p1: Point<bigint>, p2: Point<bigint>): Point<bigint> {
 /**
  * Performs a scalar multiplication by starting from the 'base' point and 'adding'
  * it to itself 'e' times.
+ * This works given the following invariant: At each step, R0 will be r_0*base where r_0 is the prefix of e
+ * written in binary and R1 will be (r_0+1)*base. In other words: at iteration i of the loop, r_0's binary
+ * representation will be the first i+1 bits of e. If the upcoming bit is a 0, we just have to double R0
+ * and add R0 to R1 to maintain the invariant. If it is a 1, we have to double R0 and add 1*base
+ * (or add R1, which is the same as (r_0+1)*base), and double R1 to maintain the invariant.
  * @param base The base point used as a starting point.
  * @param e A secret number representing the private key.
  * @returns The resulting point representing the public key.
--- a/packages/baby-jubjub/src/baby-jubjub.ts
+++ b/packages/baby-jubjub/src/baby-jubjub.ts
@@ -70,7 +70,9 @@ export function addPoint(p1: Point<bigint>, p2: Point<bigint>): Point<bigint> {
  * @returns The resulting point representing the public key.
  */
 export function mulPointEscalar(base: Point<bigint>, e: bigint): Point<bigint> {
-    console.log(e)
+    if (scalar.isZero(e)) {
+        return [BigInt(0), BigInt(1)]
+    }
     const eBits: Array<boolean> = []
     while (!scalar.isZero(e)) {
         if (scalar.isOdd(e)) {
--- a/README.md
+++ b/README.md
@@ -499,3 +499,17 @@ grep -r -l "smt" . | xargs sed -i 's/smt/my-package/'
 #### How can I create benchmarks for my library?
 
 You can see some examples in the `benchmarks` folder. All you have to do is create a file that exports a function to run your benchmark in that folder, and add that function to the `index.ts` file. The `yarn benchmarks` command can be run with no parameters (it will run all the benchmarks), or you can specify the name of your benchmark file to run just that. When you run the command it will create a `benchmarks/results` folder with your results.
+
+#### I need to use a Merkle Tree to prove the inclusion or exclusion of data elements within a set. Which type of Merkle Tree should I use?
+
+**Incremental:** Ideal for applications where you frequently add new elements and need to update the tree efficiently.
+
+**Lean Incremental:** A more memory-efficient version of the incremental Merkle tree, suitable for use cases with memory constraints.
+
+**Sparse:** Best suited for scenarios where you need to manage a large, mostly empty set of possible elements.
+
+| Type                 | Library Name     | Main Feature                           | Recommended for                          | Not Recommended for                               | Used by                                                                                                                   | Proof Gen Speed                  |
+| -------------------- | ---------------- | -------------------------------------- | ---------------------------------------- | ------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------- | -------------------------------- |
+| **Incremental**      | @zk-kit/imt      | Fastest for incremental updates        | Applications requiring frequent updates  | Scenarios with very large datasets                | [Semaphore](https://github.com/semaphore-protocol/semaphore), [Tornado Cash](https://github.com/tornadocash/tornado-core) | Fast                             |
+| **Lean incremental** | @zk-kit/lean-imt | Optimized for lightweight environments | Mobile and browser-based ZK applications | High-frequency update requirements                | [zkSync](https://github.com/matter-labs/zksync), [Loopring](https://github.com/Loopring)                                  | Moderate                         |
+| **Sparse**           | @zk-kit/smt      | Handles very large sets efficiently    | Applications with large static datasets  | Frequent updates due to higher computational cost | [Aztec](https://github.com/AztecProtocol), [Zcash](https://github.com/zcash)                                              | Slow (due to large set handling) |
--- a/packages/baby-jubjub/tests/index.test.ts
+++ b/packages/baby-jubjub/tests/index.test.ts
@@ -1,46 +1,19 @@
 import { babyjub } from "circomlibjs"
 import { utils } from "ffjavascript"
 import * as scalar from "@zk-kit/utils/scalar"
-import { Base8, Point, addPoint, inCurve, mulPointEscalar, packPoint, r, unpackPoint } from "../src"
+import { Base8, Point, addPoint, inCurve, mulPointEscalar, packPoint, r, unpackPoint, order, subOrder } from "../src"
 import { tonelliShanks } from "../src/sqrt"
 
 describe("BabyJubjub", () => {
     const secretScalar = BigInt(324)
+    const id: Point<bigint> = [BigInt(0), BigInt(1)]
 
     let publicKey: Point<bigint>
 
-    it("Test curve multiplication", async () => {
-        const P: Point<bigint> = [
-            BigInt("17777552123799933955779906779655732241715742912184938656739573121738514868268"),
-            BigInt("2626589144620713026669568689430873010625803728049924121243784502389097019475")
-        ]
-
-        expect(mulPointEscalar(P, BigInt(1)).toString()).toBe(P.toString())
-        expect(mulPointEscalar(P, BigInt(2)).toString()).toBe(addPoint(P, P).toString())
-    })
-
-    it("Test base point order", async () => {
-        const order = BigInt("21888242871839275222246405745257275088614511777268538073601725287587578984328")
-        const subOrder: bigint = BigInt("2736030358979909402780800718157159386076813972158567259200215660948447373041")
-        expect(scalar.shiftRight(order, BigInt(3))).toBe(subOrder)
-        const G: Point<bigint> = [
-            BigInt("995203441582195749578291179787384436505546430278305826713579947235728471134"),
-            BigInt("5472060717959818805561601436314318772137091100104008585924551046643952123905")
-        ]
-        const p1: Point<bigint> = mulPointEscalar(G, BigInt(8) * subOrder)
-        expect(p1[0]).toBe(BigInt(0))
-        expect(p1[1]).toBe(BigInt(1))
-        const neutral = mulPointEscalar(Base8, subOrder)
-        expect(neutral[0]).toBe(BigInt(0))
-        expect(neutral[1]).toBe(BigInt(1))
-
-        const random = BigInt("38275423985628165")
-        expect(mulPointEscalar(Base8, random)[0]).toBe(mulPointEscalar(Base8, random + BigInt(2) * subOrder)[0])
-    })
-
-    it("Test curve implementation", async () => {
-        expect(inCurve([BigInt(0), BigInt(1)])).toBeTruthy()
+    it("Test point addition and inCurve", async () => {
+        expect(inCurve(id)).toBeTruthy()
         expect(inCurve([BigInt(1), BigInt(0)])).toBeFalsy()
+        expect(addPoint(id, id).toString()).toBe(id.toString())
 
         const p1: Point<bigint> = [
             BigInt("17777552123799933955779906779655732241715742912184938656739573121738514868268"),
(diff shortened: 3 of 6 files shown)


packages/baby-jubjub/src/baby-jubjub.ts
```diff
@@ -65,25 +65,43 @@ export function addPoint(p1: Point<bigint>, p2: Point<bigint>): Point<bigint> {
 /**
  * Performs a scalar multiplication by starting from the 'base' point and 'adding'
  * it to itself 'e' times.
+ * This works given the following invariant: At each step, R0 will be r_0*base where r_0 is the prefix of e
```
artwyman, line 68: I suggest additionally referring to the named algorithm being used here (Montgomery Ladder) perhaps with a link to a description of the algorithm (Wikipedia or some other source).

packages/baby-jubjub/src/baby-jubjub.ts
```diff
@@ -65,25 +65,43 @@ export function addPoint(p1: Point<bigint>, p2: Point<bigint>): Point<bigint> {
...
-    let exp: Point<bigint> = base
-
-    while (!scalar.isZero(rem)) {
-        if (scalar.isOdd(rem)) {
-            res = addPoint(res, exp)
+    if (scalar.isZero(e)) {
+        return [BigInt(0), BigInt(1)]
+    }
+    const eBits: Array<boolean> = []
+    while (!scalar.isZero(e)) {
```
artwyman, line 82: Is the goal constant time?  This loop doesn't seem constant time to me, since it'll exit early if high bits of the input are zero.  As a result the length of eBits is also variable.  If you want constant time, it seems like there has to be a hard-coded max bit length (254 I think, but someone should double-check me) somewhere.

If you wish, I think this loop can probably be skipped entirely with some bigint bitwise operations below.  Something like this:

```ts
  for (let bitMask = 1n, bitmask < 1n<<254n, bitmask <<= 1n) {
    if (e & bitMask != 0) {
      // ...
```

I'm not sure of the relative efficiency, but I'd guess that the bitmasks are cheaper than allocating memory for an array.

packages/baby-jubjub/src/baby-jubjub.ts
```diff
@@ -65,25 +65,43 @@ export function addPoint(p1: Point<bigint>, p2: Point<bigint>): Point<bigint> {
...
  */
 export function mulPointEscalar(base: Point<bigint>, e: bigint): Point<bigint> {
-    let res: Point<bigint> = [Fr.e(BigInt(0)), Fr.e(BigInt(1))]
-    let rem: bigint = e
-    let exp: Point<bigint> = base
-
-    while (!scalar.isZero(rem)) {
-        if (scalar.isOdd(rem)) {
-            res = addPoint(res, exp)
+    if (scalar.isZero(e)) {
```
artwyman, line 78: This isn't constant time if e is zero.  Is that okay?

artwyman (commented):
I feel underqualified to review the correctness of the cryptographic algorithm, and how well it meets the goal of resistance to timing attacks.  Perhaps @cedoor is willing to do so, or can suggest someone to double-check here.  I mostly trust the unit tests elsewhere (particularly the compatibility tests in eddsa-poseidon comparing results to circomlibjs) to ensure the results of this calculation are correct.

I do feel competent to comment on some aspects of the code, and potential micro-optimizations, so I've done that here.  One thing I'm concerned about is relative performance of this algorithm compared to the old one.  Can you do some testing of the relative speed of this raw operation, and/or higher-level operations which depend on it (like eddsa-poseidon sign/verify).  Just timing operations in a loop and reporting results in the PR would be sufficient here.  I don't think this needs an automated performance test in the test suite.

cedoor:
Thanks @ChinoCribioli and @artwyman!

>  Perhaps @cedoor is willing to do so, or can suggest someone to double-check here.

I'm also not competent enough to check the logic out here. I'll ask other people to review it.

@ChinoCribioli Is this a breaking change?

packages/baby-jubjub/src/baby-jubjub.ts
chancehudson: I think although the exponent is marked as secret, it may be considered public if it's set to 0. The rationale in my mind is "no private key would ever be selected as the 0 value". That said, I don't think it's a good idea to short circuit here because it breaks the comment above of `e` being secret in _all_ cases.

packages/baby-jubjub/src/baby-jubjub.ts
chancehudson: Yes this sounds correct to me. If `e` is `7` for example this will loop 3 times, whereas if it's like `4182498124124891489144` it will loop significantly more times.

Hardcoding an iteration count is a good approach, but we'll still have timing vulnerabilities due to the use of the [javascript `BigInt`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt#cryptography) type which is not safe against timing attacks.

It's also unclear if the `scalar.isZero` function is constant time.

packages/baby-jubjub/src/baby-jubjub.ts
```diff
@@ -65,25 +65,43 @@ export function addPoint(p1: Point<bigint>, p2: Point<bigint>): Point<bigint> {
...
+        return [BigInt(0), BigInt(1)]
+    }
+    const eBits: Array<boolean> = []
+    while (!scalar.isZero(e)) {
+        if (scalar.isOdd(e)) {
```
chancehudson, line 83: Unclear if `isOdd` is constant time

packages/baby-jubjub/src/baby-jubjub.ts
```diff
@@ -65,25 +65,43 @@ export function addPoint(p1: Point<bigint>, p2: Point<bigint>): Point<bigint> {
...
+            eBits.push(true)
+        } else {
+            eBits.push(false)
         }
+        e = scalar.shiftRight(e, BigInt(1))
```
chancehudson, line 88: A right shift on a `BigInt` is almost certainly not constant time.

packages/baby-jubjub/src/baby-jubjub.ts
```diff
@@ -65,25 +65,43 @@ export function addPoint(p1: Point<bigint>, p2: Point<bigint>): Point<bigint> {
...
+    }
 
-        exp = addPoint(exp, exp)
-        rem = scalar.shiftRight(rem, BigInt(1))
+    let R0: Point<bigint> = base
+    let R1: Point<bigint> = addPoint(base, base)
+
+    for (const bit of eBits.slice(0, -1).reverse()) {
```
chancehudson, line 94: This slice is a no-op, i think. It may be necessary to invoke the `reverse` function though.

chancehudson (commented):
I'm happy to see discussion about timing vulnerabilities, and a constant time ec multiplication implementation.

Unfortunately this function is still unsafe because of the use of the javascript `BigInt` type. Internally a `BigInt` is (likely) represented as the minimum number of bytes for the value. This means any operation we do here will be variable time based on the value of the `BigInt`.

I've tried to think of how to convert from `BigInt` to an alternate representation in constant time, but haven't found an approach that seems like it would be safe. Approaches i've considered:
- use `e.toString(2)`: this will return a binary string, but it will be of variable length based on the size of the value
- use repeated division to extra le bytes: division is generally not a constant time operation based on the numerator
- do `e + (1n << 254n)` to get a ~constant size value then operate on that and discard bits above 254: this _could_ work, and may be the most safe approach, but it's unclear if the addition is constant time (the implementation may have to backfill the bits between the value and 254)

I think it's a good idea to include this implementation though (after the requested changes) to mitigate timing attacks as much as possible.

packages/baby-jubjub/src/baby-jubjub.ts
ChinoCribioli: Sure! Will do.

packages/baby-jubjub/src/baby-jubjub.ts
```diff
@@ -75,24 +76,14 @@ export function addPoint(p1: Point<bigint>, p2: Point<bigint>): Point<bigint> {
...
-    const eBits: Array<boolean> = []
-    while (!scalar.isZero(e)) {
-        if (scalar.isOdd(e)) {
-            eBits.push(true)
-        } else {
-            eBits.push(false)
-        }
-        e = scalar.shiftRight(e, BigInt(1))
-    }
+    e %= order
```
artwyman, line 79: Good thought.  I found myself wondering whether this was needed too, and whether it should be a modular reduction or an exception thrown for overly-large values.

packages/baby-jubjub/src/baby-jubjub.ts
```diff
@@ -75,24 +76,14 @@ export function addPoint(p1: Point<bigint>, p2: Point<bigint>): Point<bigint> {
...
 
-    let R0: Point<bigint> = base
-    let R1: Point<bigint> = addPoint(base, base)
+    let R0: Point<bigint> = [0n, 1n]
```
artwyman, line 81: I see there's a unit test covering multiply by zero, which is good to confirm this still behaves the same way as before without the special-case.
As a side suggestion, it might be clearer to make this a public const `id` like is used in the unit tests, to make it clear that [0n, 1n] is the representation of the zero/id point in this curve.

artwyman (commented):
Comments on the latest changes.  I'm still curious about relative performance of this implementation vs. the previous one.

packages/baby-jubjub/src/baby-jubjub.ts
ChinoCribioli: I agree with this point. I changed the loop with a hardcoded for.

packages/baby-jubjub/src/baby-jubjub.ts
ChinoCribioli: Yes, I also thought initially that we shouldn't worry too much about the zero case since it is irrelevant cryptographically. But I agree to change this to make the method more consistent.

packages/baby-jubjub/src/baby-jubjub.ts
ChinoCribioli: I think it is better to apply the reduction unless it implies another timing vulnerability, which is something I'm not sure about.

packages/baby-jubjub/src/baby-jubjub.ts
ChinoCribioli: I like that. It is done.

ChinoCribioli:
@chancehudson I just pushed most of the changes you requested. My only doubt now regarding any possible remaining timing vulnerability lies on the lines `e %= order` and `if (e & mask)`. Mostly the latter, since I don't know if the evaluation of an if statement in an integer is implemented in constant time.

ChinoCribioli:
> Comments on the latest changes. I'm still curious about relative performance of this implementation vs. the previous one.

I've run the tests of the `baby-jubjub` and `eddsa-poseidon` packages (which are the main packages affected by this PR) with both implementations and these are the results:

babyjubjub_before
babyjubjub_withMyChange
eddsa_before
eddsa_withMyChange

packages/baby-jubjub/src/baby-jubjub.ts
chancehudson:
```suggestion
    e %= order
    e += 1n << 254n
```

I suggest this so that the operation on line 88 will always operate on an `e` value of the same size. I'm not sure if the modular reduction is variable time (i'd guess yes).

chancehudson:
> on the lines `e %= order` and `if (e & mask)`. Mostly the latter, since I don't know if the evaluation of an if statement in an integer is implemented in constant time.

The if statement _should_ be fine as long as `e` is always the same size. This branch is specifically designed so that each codepath is equal time (this is the Montgomery ladder strategy). We just need to make sure the condition evaluation is constant time as well, which i suggested a change to try and ensure.

Again though, this is best effort. Regardless we're going to have _some_ timing leakage from the `BigInt` implementation.

artwyman:
> I've run the tests of the baby-jubjub and eddsa-poseidon packages (which are the main packages affected by this PR) with both implementations and these are the results:

I think this needs a more direct test of the affected operations (repeated in a loop to avoid fixed overhead and caching effects) to make the differences more clear.  It does look to me like the test times are uniformly slower, though, which isn't very pleasing.  I submitted an issue requesting to make it faster (which looks like it may be met by a WASM implementation implemented in Rust).  The utest which targets small values in particular is expected to get slower for a constant-time algorithm, but I assume the eddsa-poseidon tests are operating on more typical values.

I'm not entirely sure of the right trade-off here.  Speaking for the Zupass use case, I think we probably care more about performance than timing attacks right now, but that's not a universal choice I expect to apply to everyone.

packages/baby-jubjub/src/baby-jubjub.ts
artwyman: Am I understanding right that your proposal is to set the 254th bit to ensure uniform input size, and then intentionally ignore that bit in the for loop below?  If so, I can understand it, though it's very tricky and deserves a comment to explain it to future developers.

artwyman: A better number library, hopefully faster, would certainly be welcome, but I don't know how feasible it is.
I think there's a strategic decision to be made here which I'm not the right person to make.  How much is it worth trading speed and complexity against a partial solution to timing attacks.  There might be more full solutions to both timing and performance, but how difficult are they, and how likely are they to be accomplished soon?

chancehudson: Yes, i'm pushing for PSE to begin maintaining an open source number library with constant time operations written in Rust, but supporting webassembly. I mostly want this library for building post quantum secure constructions, but i think we could add support for things like this. Ideally we pull all this logic out of JS and into wasm where we can be much more certain of what is executing, and hopefully get equal or better performance.

This is a next couple of months type goal though, certainly not an immediate fix. @cedoor I think you're going to be the authority on this PR. I personally don't think it's worth significantly degrading performance for a fix that we're not sure solves the problem.

packages/baby-jubjub/src/baby-jubjub.ts
```diff
@@ -65,25 +67,38 @@ export function addPoint(p1: Point<bigint>, p2: Point<bigint>): Point<bigint> {
...
  */
 export function mulPointEscalar(base: Point<bigint>, e: bigint): Point<bigint> {
-    let res: Point<bigint> = [Fr.e(BigInt(0)), Fr.e(BigInt(1))]
-    let rem: bigint = e
-    let exp: Point<bigint> = base
-
-    while (!scalar.isZero(rem)) {
-        if (scalar.isOdd(rem)) {
-            res = addPoint(res, exp)
+    e %= order
```
chancehudson, line 81: Looking at this again, the original implementation doesn't make any assertions or modifications to this variable. This is an exponent, which is not necessarily a field element, so the reduction should be unnecessary.

I think if we remove this we have a _pretty safe_ function. We just need to look more deeply at the performance hit and decide if it's worth it.

artwyman: An exponent doesn't have to be a field element, but the curve is cyclic with its order, so a value greater than the order is redundant compared to the modular reduction of the same value (either leads to the same output).  And the constant-ish time algorithm below works only if you can fix the number of bits in the input.  If we reduce the modular reduction, then I think this function becomes incorrect for any value greater than the order.  Modular reduction is not the same as cutting off high bits (because the order is not a power of 2).

ChinoCribioli: Indeed. As Andrew says, some tests break when you remove the reduction because you incorrectly handle the case where the exponent passed as input is greater than 254 bits.

ChinoCribioli:
> > I've run the tests of the baby-jubjub and eddsa-poseidon packages (which are the main packages affected by this PR) with both implementations and these are the results:
> 
> I think this needs a more direct test of the affected operations (repeated in a loop to avoid fixed overhead and caching effects) to make the differences more clear. It does look to me like the test times are uniformly slower, though, which isn't very pleasing. I submitted an issue requesting to make it faster (which looks like it may be met by a WASM implementation implemented in Rust). The utest which targets small values in particular is expected to get slower for a constant-time algorithm, but I assume the eddsa-poseidon tests are operating on more typical values.
> 
> I'm not entirely sure of the right trade-off here. Speaking for the Zupass use case, I think we probably care more about performance than timing attacks right now, but that's not a universal choice I expect to apply to everyone.

I get that for some projects performance is more important than 

> > I've run the tests of the baby-jubjub and eddsa-poseidon packages (which are the main packages affected by this PR) with both implementations and these are the results:
> 
> I think this needs a more direct test of the affected operations (repeated in a loop to avoid fixed overhead and caching effects) to make the differences more clear. It does look to me like the test times are uniformly slower, though, which isn't very pleasing. I submitted an issue requesting to make it faster (which looks like it may be met by a WASM implementation implemented in Rust). The utest which targets small values in particular is expected to get slower for a constant-time algorithm, but I assume the eddsa-poseidon tests are operating on more typical values.
> 
> I'm not entirely sure of the right trade-off here. Speaking for the Zupass use case, I think we probably care more about performance than timing attacks right now, but that's not a universal choice I expect to apply to everyone.

I get that for some use cases performance is more important than safety, I honestly don't know which are the main projects that use this library. However, I think that any implementation of a rather complex cryptographic protocol must have a strong focus on security and, as you said, the performant variant may not be the best option for a lot of uses this library might have.

packages/baby-jubjub/src/baby-jubjub.ts
chancehudson: ah right because the loop count is hardcoded, i see

cedoor:
> However, I think that any implementation of a rather complex cryptographic protocol must have a strong focus on security and, as you said, the performant variant may not be the best option for a lot of uses this library might have.

I think security should have priority over performance, but I would wait for a full review by the audit team.

One solution might be to release a new major if the security benefits are substantial and continue to maintain version 1 anyway.

cedoor:
> One solution might be to release a new major if the security benefits are substantial and continue to maintain version 1 anyway.

@chancehudson @artwyman any thoughts?

artwyman:
>> One solution might be to release a new major if the security benefits are substantial and continue to maintain version 1 anyway.
> @chancehudson @artwyman any thoughts?

I have thoughts in two directions.  First general thoughts on maintaining 2 versions:

It's workable, but definitely has a cost in maintenance burden and confusion.  Either both versions need to be kept up to date with future patches, or else v1 will slowly become less useful, and potentially even insecure.  I historically have a preference for "develop on main" workflows since the code everyone looks at and uses the code which will remain up-to-date and workable.  I realize this is a preference, though, not a clear-cut definition of "best".

If you're going to support multiple algorithms for a significant time period, it may be better to keep them both in the code separately.  They could be separate packages, separate functions within the package, or configurable via an "options" parameter.  That keeps them both "on main" so they should benefit from ongoing work, unit testing, etc.

2nd, thoughts on the specifics of the security vs. performance tradeoff here:

Given bigint arithmetic isn't constant-time anyway, this algorithm isn't a complete solution to the security problem.  That makes me personally de-prioritize it vs. performance, but that's a call reasonable people can make differently.  I would defer to someone with more security expertise, and who has specific threat models in mind to make the call on whether a constant-time algorithm built on variable-time bigint math reduces timing attack risk by enough of a margin to be worth prioritizing.

Calling the current implementation the "performant variant" is admittedly a bit of a stretch, since any pure-JS implementation of cryptography isn't going to be very performant.  A preferable alternative would be to solve this problem fully as part of a more optimized implementation (probably wasm) which is fast enough to be an improvement to performance even with a constant-time algorithm.  Such an implementation could improve both performance and security at once . I'm not sure how soon that's likely to happen, though, so there may need to be trade-offs or provide multiple options in the near term.

paulmillr:
noble-curves exposes montgomery ladder here: https://github.com/paulmillr/noble-curves/blob/bee1ffe0000095f95b982a969d06baaa3dd8ce73/src/abstract/montgomery.ts#L100

No need to implement it twice

Issue #152, opened 2024-02-10
Add options to the `eddsa-poseidon` class to allow devs to pick their favorite hash function


cedoor:

hannahredler:
Hey! I'm Hannah, a dev currently studying the core program with the PSE team. I would be interested in picking this up!

My approach:
- I would choose one new algorithm to start, probably Blake2.
- I would implement this algorithm via the spec and set it as the default.
- I would add the additional ability to configure/ choose an alternative algorithm (at this point, to revert to Blake1)
- Then, in subsequent PRs, I would add the additional algorithms.

If you would be happy for me to go ahead, it would be great if you could let me know if you know of any implementations of the above algorithms that could be used, or whether it would be required to research/implement this too

cedoor:
Hey @hannahredler, great! I'll assign this task to you :)

I think a good blake2 implementation could be: https://github.com/dcposch/blakejs.

sha3: https://github.com/emn178/js-sha3.

Conditional exports should be the right way to go.

PR #329, opened 2024-09-13
feat(eddsa-poseidon): adds Blake2s hashing for eddsa and conditional imports

hannahredler:
## Description

Current implementation of the EdDSA Poseidon uses the Blake-1 hashing algorithm. This PR implements Blake2b algorithm and conditional imports such that the user can choose to use a different hashing algorithm if required.

## Related Issue(s)

Partially Closes [#152](https://github.com/privacy-scaling-explorations/zk-kit/issues/152)
Adds the first optional hashing algorithm

## Other information

The implementation of Blake2b is based on [this one](https://github.com/dcposch/blakejs), migrated to Typescript and changed to a class structure to mirror the Blake1 implementation. The test cases were lifted directly from the repo to ensure consistent output.

## Checklist

-   [ ] I have performed a self-review of my code


hannahredler:
@cedoor @artwyman the only thing I haven't managed to get to work is to import the sub packages from outside the project and check they import correctly, after running yarn build. If you could direct me in the best way to ensure this works, I can validate that

artwyman:
I don't think I have the time or expertise to review this in detail, but here are a few design discussion points to consider from the point of view of 0xPARC's projects (Zupass and PODs) making use of this library, which might be a reasonable stand-in for 

- I think it's important the packaging allow for avoiding the cost of the unused hashing options here.  E.g. if I need only blake-1, and I structure my imports appropriately, I shouldn't see the size impact of blake-2b in my bundle (and vice versa), similar to the way the poseidon-lite module works for different sizes.
- Adding more hash options definitely seems good.  Changing the default in particular is something to consider carefully, due to its impact on existing users, and the validity of their existing signed data.  It's certainly very much a breaking change, and could cause compatibility headaches for any developer who doesn't notice it's happened.  It should certainly be reflected in the semantic version, and I might go so far as to change the TypeScript interface so as to explicitly break existing code which doesn't get hand updated.  I think it would be safer to add blake-2b as an option, but leave the default as-is.

For Zupass in particular, we have existing tickets and PODs signed using the blake-1 variant so we're unlikely to switch anytime soon, and would have to introduce some new version/config fields in our formats in order to do so.  I'd expect SemaphoreV4 to be in a similar situation where they care very much about compatibility.

I'd be interested to hear the advantages of blake-2b so as to think carefully about when an upgrade might be warranted.  Are there known vulnerabilities in blake-1?  Is blake-2b stronger, faster, or superior in some other way?

hannahredler:
Thanks for the comment @artwyman.

As far as I understand it, Blake2 is optimised to be faster and more efficient, whilst reliably offering a similar level of security.  Thus if changing it could cause a headache down the line, I agree that it might be better to keep it as an option rather than changing the default, and this can always be a decision that is easily taken down the line.

In terms of bundle size, I have structured the code to only return the compiled version of the chosen algorithm, so that should be working as intended.

packages/eddsa-poseidon/src/utils.ts
```diff
@@ -71,15 +72,48 @@ export function checkMessage(message: BigNumberish): bigint {
...
+    return Array.prototype.map.call(bytes, (n) => (n < 16 ? "0" : "") + n.toString(16)).join("")
+}
+
+export function hexToBytes(hex: string) {
+    const ret = new Uint8Array(hex.length / 2)
+    for (let i = 0; i < ret.length; i += 1) {
+        ret[i] = parseInt(hex.substring(i * 2, i * 2 + 2), 16)
+    }
+    return ret
+}
```
cedoor, line 103: ZK-Kit provides a package with some utilities. 1 module can be used to convert types.

* Hex to buffer: https://github.com/privacy-scaling-explorations/zk-kit/blob/2173bc800d2a22409678a7d3983df48d4f8efd18/packages/utils/src/conversions.ts#L201
* Buffer to hex: https://github.com/privacy-scaling-explorations/zk-kit/blob/2173bc800d2a22409678a7d3983df48d4f8efd18/packages/utils/src/conversions.ts#L218

Also, if a function is only used in tests, it shouldn't be part of the source files.

cedoor (changes requested):
Hey @hannahredler, thanks for this PR!

I added some comments 👍🏽

packages/eddsa-poseidon/src/blake2b.ts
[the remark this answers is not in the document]
hannahredler: That makes sense. Ive modified the Blake2b class to wrap the existing library 👍

packages/eddsa-poseidon/src/HashFunction.ts
```diff
@@ -0,0 +1,4 @@
+export interface HashFunction {
```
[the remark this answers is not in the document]
hannahredler: Blake and Blake2b  classes were supposed to be implementing it, updated now

packages/eddsa-poseidon/src/utils.ts
```diff
@@ -71,15 +72,48 @@ export function checkMessage(message: BigNumberish): bigint {
     return bufferToBigInt(Buffer.from(message as string))
 }
 
+const ERROR_MSG_INPUT = "Input must be an string, Buffer or Uint8Array"
+
+// For convenience, let people hash a string, not just a Uint8Array
+export function normalizeInput(input: string | Uint8Array) {
```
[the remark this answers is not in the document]
hannahredler: Okay makes sense, done 👍

packages/eddsa-poseidon/src/utils.ts
hannahredler: Thanks! Removed them from utils and using the hexadecimalToBuffer function from zk-kit instead

packages/eddsa-poseidon/src/index.ts
```diff
@@ -1,2 +1 @@
-export * from "./eddsa-poseidon"
-export * from "./types"
+export * from "./eddsa-poseidon-blake-2b"
```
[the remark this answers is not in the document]
hannahredler: Changed the default export back to blake 1 as suggested

packages/eddsa-poseidon/src/index.ts
```diff
@@ -1,2 +1 @@
-export * from "./eddsa-poseidon"
-export * from "./types"
```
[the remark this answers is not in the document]
hannahredler: They haven't been removed, they're actually just being exported from the inner file to avoid duplication. Happy for any feedback on the export structure

packages/eddsa-poseidon/src/utils.ts
```diff
@@ -71,15 +72,48 @@ export function checkMessage(message: BigNumberish): bigint {
...
+    const ret = new Uint8Array(hex.length / 2)
+    for (let i = 0; i < ret.length; i += 1) {
+        ret[i] = parseInt(hex.substring(i * 2, i * 2 + 2), 16)
+    }
+    return ret
+}
+
+// TODO: Remove?
 /**
  * Computes the Blake512 hash of the input message.
```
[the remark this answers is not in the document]
hannahredler: done 👍

packages/eddsa-poseidon/src/HashFunction.ts
```diff
@@ -0,0 +1,4 @@
...
+    update(data: Buffer): HashFunction
+    digest(): Buffer
```
cedoor, line 3: Not sure if this is necessary, but it could be better to use the type provided by the `buffer` package here, instead of the native nodejs Buffer type. You should just add an import.

packages/eddsa-poseidon/src/index.ts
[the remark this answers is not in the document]
cedoor: Oh yea, that should work 👍🏽

packages/eddsa-poseidon/src/blake.ts
```diff
@@ -156,7 +157,7 @@ function lengthCarry(arr: number[]) {
  * hashing, allowing data to be added in chunks.
  */
 /* eslint-disable import/prefer-default-export */
-export class Blake512 {
+export class Blake512 implements HashFunction {
```
cedoor, line 160: Let's make `Blake512` and `Blake2b` class exports consistent. Like making this the default import.

cedoor (changes requested):

packages/eddsa-poseidon/src/blake.ts
hannahredler: done!

packages/eddsa-poseidon/src/HashFunction.ts
hannahredler: also done

packages/eddsa-poseidon/src/utils.ts
```diff
@@ -72,14 +74,15 @@ export function checkMessage(message: BigNumberish): bigint {
...
  * @param message The input data to hash, provided as a Buffer.
- * @returns A Buffer containing the 512-bit hash result.
+ * @param algorithm The selected algorithm
+ * @returns A Buffer containing the result
  */
-export function hash(message: Buffer | Uint8Array): Buffer {
-    const engine = new Blake512()
+export function hashInput(message: Buffer | Uint8Array, algorithm?: SupportedHashingAlgorithms) {
+    let engine: HashFunction
+    if (!algorithm || algorithm === SupportedHashingAlgorithms.BLAKE1) engine = new Blake512()
```
cedoor, line 84: If I pass an unsupported algorithm, this function would still return Blake2b right? I think it's better to use a `switch` here, where the default case throws an exception warning devs that the algorithm is not supported.

packages/eddsa-poseidon/tests/blake2.test.ts
cedoor, line 1: Could you rename this file to `eddsa-poseidon-blake2b.test.ts`?

cedoor (changes requested):
LGTM, just two comments before merging!

packages/eddsa-poseidon/tests/blake2.test.ts
hannahredler: done

packages/eddsa-poseidon/src/utils.ts
hannahredler: good point, done!

PR #353, opened 2024-11-26
Fix for loop structure to improve readability

0xwitty:
## Description

This update refactors the loop to use a more concise arrow function, which improves the clarity of the code. Previously, the `numberOfLeaves` variable was defined with a broader scope, but by using a `const` inside the loop, its scope is now limited to the loop's iteration. This small change not only enhances code readability but also helps prevent potential issues related to variable scope, making the code easier to maintain in the future.

## Checklist

### DIFF
--- a/benchmarks/index.ts
+++ b/benchmarks/index.ts
@@ -3,6 +3,7 @@ import run from "./benchmark-merkle-trees"
 const treeDepthSamples: number[] = [3, 7, 10]
 let numberOfLeaves: number
 treeDepthSamples.forEach((treeDepth) => {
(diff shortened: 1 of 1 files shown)


0xDatapunk (approved):
nice

cedoor (commented):
Hi @0xwitty!  What's the point of keeping  `let numberOfLeaves: number`? Can we remove it?

sripwoud (commented):
If we start refactoring...
I think the logic if simple enough to avoid declaring  `numberOfLeaves` altogether

PR #354, opened 2024-11-29
feat(smt): allow ChildNodes to use 0 as entry marker

signorecello:
It's the nittiest of nits, but helps a lot if for whatever reason you want to use 0 as entry marker for the leaves in an SMT


cedoor:
Oh makes sense! Thanks @signorecello

PR #355, opened 2024-12-02
fix(lean-imt): After a sequence of updates a proof is invalid.

Lauman:
## Description

When a member is removed from the tree (update a leaf with 0n) and then another member is updated, the root that is saved is not correct, so the proof that is generated is not valid.This was caused by a type validation failure.
This fix the behavior described above.
I detected this when I was testing in Semphore groups.

## Checklist


0xDatapunk (commented):
what is the expected behavior of poseidon, when one of the inputs is undefined?

I am wondering if this should be fixed there instead

Lauman:
> what is the expected behavior of poseidon, when one of the inputs is undefined?
> 
> I am wondering if this should be fixed there instead

Hi 0xDatapunk. The poseidon method accepts bigint|number|string, if you pass an undefined parameter the process crashes with the error TypeError: Cannot convert undefined to a BigInt.
I think if there is no data for that sibiling it is not necessary to perform the hash.

cedoor (approved):

Lauman:
@cedoor I had to rebase my commit because they weren't verified. I've fixed it now.

cedoor (approved):

cedoor (approved):

PR #361, opened 2024-12-12
fix(eddsa-poseidon): fixes carry handling of g function by introducing wrapper over original implementation

### LINKED ISSUE #344: Fix carry handling in the g function in blake.ts
## Incorrect Carry Handling in the `g` Function
The `g` function in Implementation in [`blake.ts`](https://github.com/privacy-scaling-explorations/zk-kit/blob/main/packages/eddsa-poseidon/src/blake.ts#L101) uses `~~(lo / 0x0100000000)` to compute the carry from the lower 32 bits of a 64-bit word. 

Since `lo` can be up to `0x2FFFFFFFC` (i.e., approximately 3 times `0x0100000000`), the carry can erroneously be 2 or 3.

**Impact**
- Functional Integrity: Incorrect carry values can corrupt the internal state, leading to wrong hash outputs.
- Security Risks: The integrity of the hash function is compromised, potentially allowing for hash collisions or predictable outputs, which undermines the cryptographic strength of Blake2-512.

**Recommendation**
- Modify the carry calculation to ensure that only a single carry bit (0 or 1) is propagated. For example:

```typescript
const carry = lo >= 0x100000000 ? 1 : 0;
v[a * 2] = (v[a * 2] + ((m[sigma[i][e] * 2] ^ u512[sigma[i][e + 1] * 2]) >>> 0) + v[b * 2] + carry) >>> 0;
```
- Alternatively, use BigInt for precise 64-bit arithmetic operations as in the original Blake implementation in the [npm repo](https://www.npmjs.com/package/blake-hash), which TypeScript supports, to handle carries correctly without manual intervention.

Arch0125:
## Description

- This PR replaces the existing blake-hash implementation with a TS wrapper over the original blake-hash [npm package](https://github.com/cryptocoinjs/blake-hash) 
- Also includes required types for the blake-hash repo which was missing

## Related Issue(s)

Closes #344 

## Checklist

### DIFF
--- a/packages/eddsa-poseidon/package.json
+++ b/packages/eddsa-poseidon/package.json
@@ -60,6 +60,7 @@
     "dependencies": {
         "@zk-kit/baby-jubjub": "1.0.3",
         "@zk-kit/utils": "1.2.1",
+        "blake-hash": "^2.0.0",
         "blakejs": "^1.2.1",
         "buffer": "6.0.3",
         "poseidon-lite": "0.3.0"
--- a/packages/poseidon-cipher/src/types/blake-hash.d.ts
+++ b/packages/poseidon-cipher/src/types/blake-hash.d.ts
@@ -1,0 +1,15 @@
+declare module "blake-hash" {
+    type BlakeVariant = "blake224" | "blake256" | "blake384" | "blake512"
+
+    export interface BlakeHash {
+        update(data: Buffer | Uint8Array): BlakeHash
+        digest(encoding?: "hex" | "binary"): Buffer | string
+    }
+
+    /**
+     * Creates an instance of a BLAKE hash function.
+     * @param variant The desired BLAKE variant ('blake224', 'blake256', 'blake384', or 'blake512').
+     * @returns An instance of the hash function.
+     */
+    export default function createBlakeHash(variant: BlakeVariant): BlakeHash
+}
--- a/packages/eddsa-poseidon/src/blake.ts
+++ b/packages/eddsa-poseidon/src/blake.ts
@@ -14,277 +14,39 @@
  * hashes of input data, providing both one-time hashing capabilities and incremental
  * hashing to process large or streaming data.
  *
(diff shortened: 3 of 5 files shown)


packages/eddsa-poseidon/package.json
```diff
@@ -60,6 +60,7 @@
     "dependencies": {
         "@zk-kit/baby-jubjub": "1.0.3",
         "@zk-kit/utils": "1.2.1",
+        "blake-hash": "^2.0.0",
         "blakejs": "^1.2.1",
```
artwyman, line 64: - Does the blakejs library provide a blake1 hash which could be used here instead?
- Alternately, have you checked whether the blakejs library might have the same error that this PR is fixing, and need its own update for the blake2b use case?

artwyman (commented):
I'll defer to @cedoor to drive the real review here with more context.
For my thoughts/concerns, see my commentary in the related issue here: https://github.com/privacy-scaling-explorations/zk-kit/issues/344#issuecomment-2539930148

packages/poseidon-cipher/src/types/blake-hash.d.ts
cedoor, line 1: Why this type in the `poseidon-chipher` lib?

packages/eddsa-poseidon/package.json
cedoor: > Does the blakejs library provide a blake1 hash which could be used here instead?

`blakejs` should only provide the following implementations: BLAKE2b and BLAKE2s.

> Alternately, have you checked whether the blakejs library might have the same error that this PR is fixing, and need its own update for the blake2b use case?

Let me tag @thogiti for this question.

cedoor (changes requested):
Following up on the @artwyman's comment: https://github.com/privacy-scaling-explorations/zk-kit/issues/344#issuecomment-2539930148. 

It would be nice to have a test for this fix.

PR #365, opened 2025-01-12
fix: Fix regex issue Update rollup.config.ts

ursulabauer:
## Description

I noticed a bug in the regex used for transforming package names. The current code attempts to access `x[1]` in the replacement function, which can be `undefined` if `x` is a single character (like `-` or `/`). This causes the script to fail.  

I’ve fixed it by using `x.slice(1)` instead, which safely handles single-character cases. Here's the corrected line:  

```javascript  
const name = pkg.name.split("/")[1].replace(/[-/]./g, (x: string) => x.slice(1).toUpperCase())  
```  

This change ensures the script works as expected without errors.

## Checklist

### DIFF
--- a/packages/baby-jubjub/rollup.config.ts
+++ b/packages/baby-jubjub/rollup.config.ts
@@ -16,7 +16,7 @@ const banner = `/**
  * @see [Github]{@link ${pkg.homepage}}
 */`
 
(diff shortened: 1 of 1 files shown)


cedoor:
Hi @ursulabauer, package names in this repo should always follow the following format: `@zk-kit/<package-name>`. So it's fine if that script fails, as it means that the name is not supported.

PR #420, opened 2026-06-07
fix(smt): reject forged membership proofs on an empty tree

### LINKED ISSUE #343: verifyProof probably should return false for any key if the tree is empty
https://github.com/privacy-scaling-explorations/zk-kit/blob/253277c91c8ffee33bfb4bc5abd8a0db0251df61/packages/smt/src/smt.ts#L237

not sure if it is as designed, but verifyProof return True for any key with an empty tree. It makes better sense if false is returned:
```
verifyProof(merkleProof: MerkleProof): boolean {
    if (this.root === this.zeroNode) {
        // Empty tree case
        return merkleProof.root === this.zeroNode && 
               merkleProof.siblings.length === 0 &&
               merkleProof.matchingEntry === undefined &&
               merkleProof.membership === false;
    }
    // ... rest of the method ...
}```

### PR DESCRIPTION
**Pull request by zkasuran**
## Description

Closes #343.

`SMT.verifyProof` recomputes the root from the proof's own siblings and compares it **only** against the root stored inside the proof (`merkleProof.root`), never against the tree's state. On an empty tree this lets a self-consistent, fabricated proof pass as a **membership** proof:

```ts
const tree = new SMT(hash) // empty, root === zeroNode
const entry = ["1", "2", "1"]
tree.verifyProof({
  entry,
  matchingEntry: undefined,
  siblings: [],
  root: hash(entry), // fabricated
  membership: true
}) // => true  ❌  (no key is actually a member of an empty tree)
```

It hits the no-matching-entry branch where `calculateRoot(hash(entry), path, [])` trivially equals the fabricated `merkleProof.root`, so the check passes.

## Fix

Add an early guard in `verifyProof`: when the tree is empty (`this.root === this.zeroNode`), the only valid proof is a **non-membership** proof for the zero root with no siblings and no matching entry.

```ts
if (this.root === this.zeroNode) {
    return (
        merkleProof.root === this.zeroNode &&
        merkleProof.siblings.length === 0 &&
        merkleProof.matchingEntry === undefined &&
        merkleProof.membership === false
    )
}
```

## On the discussion in the issue

- @cedoor — _"A proof of non-membership in an empty tree should still be `true`."_ ✅ It is. A genuine non-membership proof from `createProof` on an empty tree is `{ root: zeroNode, siblings: [], matchingEntry: undefined, membership: false }`, which satisfies the guard and still verifies. The added `Should verify a non-membership proof against an empty tree` test covers this.
- @cedoor — _"Not sure about `matchingEntry === undefined`, should a `matchingEntry` still be defined?"_ On an empty tree it cannot be defined: `retrieveEntry`'s descent loop runs only while `node !== this.zeroNode`, so on an empty tree it returns immediately with `{ entry: [key], siblings: [] }` and no `matchingEntry`. A defined `matchingEntry` against an empty tree therefore only ever comes from a forged proof, so rejecting it is correct.

## Tests

Two `it.each` cases over `sha256` / `poseidon2` / `pedersen`:
- `Should not verify a forged membership proof against an empty tree` — fails before the fix, passes after.
- `Should verify a non-membership proof against an empty tree` — guards against false rejection.

`yarn test:library smt` → 63 passing, 100% coverage (statements / branches / functions / lines). `eslint` and `prettier` clean.

## Scope

This PR addresses the empty-tree case reported in #343. The more general observation that `verifyProof` validates only internal proof consistency (a fabricated `proof.root` can also be self-consistent on a **non-empty** tree) is a separate, broader change and is intentionally out of scope here; happy to open a follow-up issue if useful.

### DIFF
--- a/packages/smt/src/smt.ts
+++ b/packages/smt/src/smt.ts
@@ -235,6 +235,19 @@ export default class SMT {
      * @returns True if the proof is valid, false otherwise.
      */
     verifyProof(merkleProof: MerkleProof): boolean {
+        // If the tree is empty the only valid proof is a non-membership proof
+        // for the zero root with no siblings. This prevents a self-consistent
+        // proof built around a fabricated root from being accepted as a valid
+        // membership proof against an empty tree.
+        if (this.root === this.zeroNode) {
+            return (
+                merkleProof.root === this.zeroNode &&
+                merkleProof.siblings.length === 0 &&
+                merkleProof.matchingEntry === undefined &&
+                merkleProof.membership === false
+            )
+        }
+
         // If there is not a matching entry it simply obtains the root
         // hash by using the siblings and the path of the key.
         if (!merkleProof.matchingEntry) {
--- a/packages/smt/tests/index.test.ts
+++ b/packages/smt/tests/index.test.ts
@@ -225,6 +225,40 @@ describe("SMT", () => {
 
             expect(tree.verifyProof(proof)).toBeFalsy()
         })
+
+        it.each([["sha256"], ["poseidon2"], ["pedersen"]])(
+            "Should not verify a forged membership proof against an empty tree - %s",
+            (hash) => {
+                const hashFunction = hashes[hash as keyof typeof hashes]
+                const tree = new SMT(hashFunction)
+
+                // The tree is empty, so its root is the zero node and no key is a
+                // member. A self-consistent proof whose own root matches a fabricated
+                // leaf must not be accepted as a membership proof for this tree.
+                const entry: ChildNodes = ["1", "2", "1"]
+                const forgedProof = {
+                    entry,
+                    matchingEntry: undefined,
+                    siblings: [],
+                    root: hashFunction(entry),
+                    membership: true
+                }
+
+                expect(tree.verifyProof(forgedProof)).toBeFalsy()
+            }
+        )
+
+        it.each([["sha256"], ["poseidon2"], ["pedersen"]])(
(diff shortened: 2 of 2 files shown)


cedoor:
Hey @zkasuran, I'd defer this to @vplasencia as I'm no longer working on this project.

zkasuran:
@vplasencia cedoor handed this over to you back in June so I am bringing it to you directly.

On current `main` (f5bfea8) a proof carrying a fabricated root still verifies as a membership proof against an empty tree, so `verifyProof` returns `true` for a key no tree holds. Any caller using `verifyProof` as an inclusion check where the tree can be empty accepts that forged proof.

The test is `Should not verify a forged membership proof against an empty tree` in `packages/smt/tests/index.test.ts`. I applied that test file alone to current `main` with the source untouched: it fails 3 of 3 cases (sha256, poseidon2, pedersen), each returning `true` where `false` is expected. With the guard in `verifyProof` the smt suite is 63/63 at 100% statements, branches, functions and lines. `eslint` and `prettier -c` are clean.

I merged `main` into the branch just now (27e6143). It is current main plus 47 added lines and 0 deleted, and none of the 9 commits since I opened this touch `packages/smt`.

On the BLOCKED label, it is the pending review and nothing else. CI has also never run here. The old head sha had 0 workflow runs and 0 check runs, and the push above created one that came straight back as `action_required`: https://github.com/zk-kit/zk-kit/actions/runs/30738735414. That is the fork workflow approval gate, so this PR has never had a green or a red signal on it for anyone to judge it by. Approving that run produces one.

PR #421, opened 2026-07-08
fix(lean-imt): fixed zero values bug for updateMany and insertMany

### LINKED ISSUE #419: Bug: LeanIMT zero leafs break insertMany and updateMany
**Describe the bug**
LeanIMT assumes a leaf that is 0 to not exist. And creates invalid root. 
At insertMany at line [191](https://github.com/zk-kit/zk-kit/blob/6f7987cf783ca4915ff7f9311091b244ae62210f/packages/lean-imt/src/lean-imt.ts#L191)   
https://github.com/zk-kit/zk-kit/blob/6f7987cf783ca4915ff7f9311091b244ae62210f/packages/lean-imt/src/lean-imt.ts#L191
At updateMany at line [285](https://github.com/zk-kit/zk-kit/blob/6f7987cf783ca4915ff7f9311091b244ae62210f/packages/lean-imt/src/lean-imt.ts#285)   
https://github.com/zk-kit/zk-kit/blob/6f7987cf783ca4915ff7f9311091b244ae62210f/packages/lean-imt/src/lean-imt.ts#L285

**To Reproduce**
insert zero in the tree that lands at an uneven index (so it's an right node)
Then update or insert a leaf  with insertMany or updateMany. In such a way the new leaf lands to the left of that zero.   
Code ignores that zero leaf to the right of the new/updated leafs. Root become invalid.

**example**
```js
const tree = new LeanIMT(hash)
tree.insertMany([1n, 0n])
// root becomes 1n, should be hash(1n, 0n)
```

**Expected behavior**
code to treat 0n children as real nodes. Just like in the fix done for update. Where it check `child !== undefined`
https://github.com/zk-kit/zk-kit/blob/6f7987cf783ca4915ff7f9311091b244ae62210f/packages/lean-imt/src/lean-imt.ts#L228

**fix**
Simply !== undefined on these :D

jimjimvalkema:
## Description
InsertMany and updateMany do `node ?`  instead of  `node !== undefined ?`. This causes 0 values not to written to this._nodes, which silently causes invalid roots. 

The fix was doing `node !== undefined` instead of just `node ?` for both updateMany and insertMany.  
I also added two small test that expose this bug (which is now fixed).

## Related Issue(s)

Fixes #419 

## Checklist

-   [ x] I have read and understand the [contributor guidelines](https://github.com/zk-kit/zk-kit/blob/main/CONTRIBUTING.md) and [code of conduct](https://github.com/zk-kit/zk-kit/blob/main/CODE_OF_CONDUCT.md).
-   [ x] I have performed a self-review of my code
-   [ x] I have commented my code, particularly in hard-to-understand areas
-   [ x] My changes generate no new warnings
-   [ x] I have run `yarn style` without getting any errors
-   [ x] I have added tests that prove my fix is effective or that my feature works
-   [ x] New and existing unit tests pass locally with my changes

### DIFF
--- a/packages/lean-imt/src/lean-imt.ts
+++ b/packages/lean-imt/src/lean-imt.ts
@@ -188,7 +188,7 @@ export default class LeanIMT<N = bigint> {
                 const rightNode = this._nodes[level][index * 2 + 1]
                 const leftNode = this._nodes[level][index * 2]
 
-                const parentNode = rightNode ? this._hash(leftNode, rightNode) : leftNode
+                const parentNode = rightNode !== undefined ? this._hash(leftNode, rightNode) : leftNode
 
                 this._nodes[level + 1][index] = parentNode
             }
@@ -282,7 +282,7 @@ export default class LeanIMT<N = bigint> {
             for (const index of modifiedIndices) {
                 const leftChild = this._nodes[level - 1][2 * index]
                 const rightChild = this._nodes[level - 1][2 * index + 1]
-                this._nodes[level][index] = rightChild ? this._hash(leftChild, rightChild) : leftChild
+                this._nodes[level][index] = rightChild !== undefined ? this._hash(leftChild, rightChild) : leftChild
                 newModifiedIndices.push(index >> 1)
             }
             modifiedIndices = new Set<number>(newModifiedIndices)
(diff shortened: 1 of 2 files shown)


vplasencia (approved):
Looks great @jimjimvalkema! Thank you very much for creating the issue and the PR with all the detailed information 🙏

